PR #13, opened 2019-03-19
[keyman.com] Add note for kmcomp in download page

### LINKED ISSUE #12: [keyman.com] Keyman Developer compiler download needs some information
On this page: https://keyman.com/developer/download.php
We have the option to download the compiler, but nothing is said about it or why you would need to download it.

A little explanation on the page would be helpful.

Per Doug Higby

mcdurdin:
Fixes #12

### DIFF
--- a/developer/download.php
+++ b/developer/download.php
@@ -24,5 +24,13 @@
 
 <?php
   downloadLargeCTA('Keyman Developer', 'developer', 'stable', 'keymandeveloper-$version.exe');
+?>
+
+<h2 class='red underline'>kmcomp Compiler</h2>
+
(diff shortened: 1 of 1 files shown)


developer/download.php
```diff
@@ -24,5 +24,13 @@
 
 <?php
   downloadLargeCTA('Keyman Developer', 'developer', 'stable', 'keymandeveloper-$version.exe');
+?>
+
+<h2 class='red underline'>kmcomp Compiler</h2>
+
+<p>The kmcomp compiler is a command line tool that can be run on Windows, or using WINE on Linux and macOS. <a href='https://help.keyman.com/developer/10.0/context/kmcomp'>kmcomp reference</a>. This download is
```
darcywong00: Can we reference 11.0?

PR #21, opened 2019-05-30
[keyman.com] After recent changes to root web.config, some redirect rules were failing to apply.

mcdurdin:
Fixes keymanapp/keyman#1806


mcdurdin:
Overriding code review and merging

PR #24, opened 2019-08-06
Add beta announcement

mcdurdin:


mcdurdin:
The outside contribution was not included in the 12.0 release notes because
we back-ported it to 11.0, so it's not new in 12.0.

PR #25, opened 2019-08-15
Fix redirects to beta link

darcywong00:
During beta release, the "Learn more" links on the products pages (e.g. https://keyman.com/linux/) are pointing to `$beta_version` **/12.0**, but the content is `$beta_version_int` **/12**.

### DIFF
--- a/android/index.php
+++ b/android/index.php
@@ -22,7 +22,7 @@
 <?php
     if (betaTier()) {
 ?>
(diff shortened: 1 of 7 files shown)


mcdurdin:
We had a redirect in /web.config for `11.0` -> `11`. We should do the same for `12.0` and so on:

```
                <rule name="/11.0 to /11 landing page" stopProcessing="true">
                    <match url="^11(\.0)\/?$" />
                    <action type="Redirect" url="11/" />
                </rule>
```

darcywong00:
> We should do the same for 12.0 and so on:

Ah, I missed that when I skimmed through web.config.

PR #26, opened 2019-09-17
Updates iOS product page with 12.0 info

jahorton:


darcywong00:
For previous stable releases, we'd put the content on a staging-xx branch and only merge it after the official release is live.

help.keyman.com is allowed to have "future" content, but
keyman.com is a front-facing site for the stable version.

jahorton:
OK.  Well, consider this ready in advance for the 'staging' branch, then.

darcywong00:
How about editing this PR to merge into `staging-12.0` instead of `master`? Then we can pile onto that branch.

jahorton:
Oh.  Didn't realize it already existed.

PR #46, opened 2019-10-25
Add linux to front page core products

darcywong00:
Adds Keyman for Linux to the front page.

Question: Do we want to remove the bookmarklet panel to keep the 2x3?


mcdurdin:
> Question: Do we want to remove the bookmarklet panel to keep the 2x3?

No -- let's go with 7 boxes

PR #47, opened 2019-10-26
Collect analytics on keyboard downloads and trigger build of keyboard bundle when keyboard download is started

mcdurdin:
Part of https://github.com/keymanapp/keyman/issues/2248

Initial work on this feature. This redirects all keyboard downloads through a proxy page to allow for triggering of a rebuild of the bundled installer on demand. As a side benefit, implemented the keyboard analytics.

Still to do:
- [x] ~~Add support for triggering an initial version of a bundled installer when it is missing (it will happen if a user selects the .kmp but it's a little less obvious)~~
- [x] Consider moving the trigger and GA callbacks into an asynchronous response (the user doesn't need to wait for those to complete)
- [x] Control for multiple users downloading the same file avoiding multiple build triggers

### DIFF
--- a/_includes/async/keyboard-download.php
+++ b/_includes/async/keyboard-download.php
@@ -1,5 +1,5 @@
 <?php
-  define('DEBUG', 1); // 0 = off, 1 = lots of logging but don't force a build, 2 = force a build
+  define('DEBUG', 0); // 0 = off, 1 = lots of logging but don't force a build, 2 = force a build
 
   if(!is_cli()) {
     echo 'keyboard-download.php must be run from the command line, not web server environment.';
@@ -32,6 +32,8 @@
   recordGoogleAnalyticsEvent($cid, $id, $platform, $mode);
 
   if(hasNewerBundleVersion($id, $downloads)) {
+    // TeamCity by default will not queue multiple instances of builds with similar properties
+    // so it doesn't hurt to re-trigger it.
     triggerBuildOfBundle($id);
   }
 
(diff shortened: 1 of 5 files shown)


mcdurdin:
> * Add support for triggering an initial version of a bundled installer when it is missing (it will happen if a user selects the .kmp but it's a little less obvious)

Moving this into a separate PR. Design wise, it will trigger a slightly different route because there will be no download afterward. There will need to be some front end changes to show a box for the bundled installer and not cause confusion when it isn't available.

_includes/async/keyboard-download.php
```diff
@@ -0,0 +1,200 @@
...
+      $keyboardVersion = $keyboard_info->version;
+      $supportsWindows = is_object($keyboard_info) && isset($keyboard_info->platformSupport->windows) && $keyboard_info->platformSupport->windows != 'none';
+    } else {
+      // We avoid
+      error_log("Failed to download $url: $php_errormsg");
+      $supportsWindows = false;
+    }
+
+    if($supportsWindows) {
+      if(!isset($downloads->windows) && $target == 'windows') {
```
darcywong00: Where is `$target` set?

keyboards/download.php
```diff
@@ -11,7 +11,7 @@
   //   mode:     optional, either bundle or standalone (for now, supported only for Windows)
   //   cid:      optional, adds client id for Google Analytics tracking
```
darcywong00: Aren't the four parameters required now? (in the order `$cid, $id, $platform, $mode`)

darcywong00: nvm, I see the default values in `validateParameters()`

_includes/async/keyboard-download.php
mcdurdin: Good catch; will fix!

mcdurdin: Fixed.

PR #48, opened 2019-11-07
fix: bundled installer trigger was not running

mcdurdin:
* The `$_ENV` superglobal was not populated on azure so we did not see the variable. Also renamed and made all caps, probably not necessary but not important.
* The download trigger went to the keyboard id, not the keyboard source path as is required by the build scripts.

### DIFF
--- a/_includes/async/keyboard-download.php
+++ b/_includes/async/keyboard-download.php
@@ -8,8 +8,8 @@
 
   require_once(dirname(__FILE__)."\\..\\includes\\servervars.php");
 
-  if($argv < 4) {
-    error_log('Parameters: cid id platform mode');
+  if($argv < 5) {
+    error_log('Parameters: cid id platform mode bearer_token(teamcity)');
     exit(1);
   }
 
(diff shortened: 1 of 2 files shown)


_includes/async/keyboard-download.php
```diff
@@ -150,25 +151,43 @@ function hasNewerBundleVersion($id, $downloads) {
     return $shouldBuild;
   }
 
+  /**
+   * Retrieves the keyboard source path from the keyboard_info api
+   */
+  function getKeyboardSourcePath($id) {
+    // Get the source path
+    $keyboard_info = @file_get_contents("https://api.keyman.com/keyboard/$id");
+    if($keyboard_info === FALSE) {
```
darcywong00: If a keyboard isn't found, won't $keyboard_info = 
```
{
    "message": "Keyboard not found"
}
```

darcywong00: nvm, I see the status FALSE

keyboards/download.php
```diff
@@ -81,11 +81,18 @@ function validateParameters(&$id, &$platform, &$mode, &$cid, &$target) {
    * Start a background PHP process to do the async work so we don't block the user
    */
   function triggerDownloadBackgroundProcesses($cid, $id, $platform, $mode) {
+    $bearer_token = getenv('TEAMCITY_TOKEN');
+    if($bearer_token === FALSE) {
+      error_log("ERROR: [download.php] TEAMCITY_TOKEN is not configured.");
```
darcywong00: Do you want this in a block?
```
if(DEBUG) {
  error_log("ERROR: ...");
}
```

mcdurdin: This one is a configuration problem so I would like to be able to see it in the logs if it fails. The others may well be bad input so they just add noise to the log.

PR #51, opened 2019-11-13
fix: embed downloads are not internal

mcdurdin:
The embedded mode of keyboard downloads was subtly broken after the
unbundled installer changes. Instead of using a wrapped link pointing
to keyman:download?..., the browser's default download handler would
be triggered.

Fixes keymanapp/keyman#2332.

### DIFF
--- a/_includes/includes/ui/keyboard-details.php
+++ b/_includes/includes/ui/keyboard-details.php
@@ -34,7 +34,7 @@
 
     static private $deprecatedBy;
 
-    private function WriteLinkAnnotator() {
+    private static function WriteLinkAnnotator() {
       echo '<script>
         var binaryFileClientId = null;
         function downloadBinaryFile(a) {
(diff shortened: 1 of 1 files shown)


_includes/includes/ui/keyboard-details.php
```diff
@@ -90,16 +90,28 @@ public static function get_tier_from_request($tier = 'stable') {
...
       $id = self::$id;
-      $downloadlink = "<a class='download-link binary-download' href='download?id=$id&platform=$platform&mode=$mode' onclick='return downloadBinaryFile(this);'><span>$linktitle</span></a>";
+      global $embed;
+      
+      $e_filename = urlencode($filename);
+      $e_id = urlencode($id);
+
+      if($embed != 'none') {
+        // note: if embed != none, mode should was be standalone
+        $url = "keyman:download?filename=$e_filename&url=".urlencode("https://keyman.com/keyboard/download?id=$e_id&platform=$platform&mode=$mode");
```
darcywong00: Looking down the road when the mobile apps will browser/download keyboards through these links, will the apps need to handle `keyman:download`?

mcdurdin: Possibly; we'll cross that design bridge when we come to the update.

PR #52, opened 2019-11-18
feat: training home page

mcdurdin:
Links to pre-registration form, will need to be refreshed
once we have a full registration form up.

### DIFF
--- a/index.php
+++ b/index.php
@@ -19,7 +19,7 @@
     
         <div class="section section-announcement">
           <div class='wrapper'>
-            <p><a href='/12/'>New: Keyman 12.0 is now available!</a></p>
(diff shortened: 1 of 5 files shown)


index.php
```diff
@@ -19,7 +19,7 @@
     
         <div class="section section-announcement">
           <div class='wrapper'>
-            <p><a href='/12/'>New: Keyman 12.0 is now available!</a></p>
+            <p style='font-size:2em'><a href='/training/'>🎓 New! Keyman training in Switzerland, Feb 2020</a></p>
```
darcywong00: In a medium viewport (between ~846px to ~1080px wide), the title is getting covered up
image

PR #62, opened 2019-12-07
fix: hide deprecated keyboards by default

mcdurdin:
This depends on an API update in keymanapp/api.keyman.com#28.

### DIFF
--- a/_includes/includes/ui/keyboard-details.php
+++ b/_includes/includes/ui/keyboard-details.php
@@ -273,8 +273,8 @@ END;
 
     protected static function WriteiPadBoxes() {
       global $appstore;
-      if (isset(self::$downloads->js)) {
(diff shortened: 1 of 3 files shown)


_includes/includes/ui/keyboard-details.php
```diff
@@ -274,7 +274,7 @@ protected static function WriteiPhoneBoxes() {
     protected static function WriteiPadBoxes() {
       global $appstore;
       if (isset(self::$downloads->js)) {
```
darcywong00: Why does WriteiPadBoxes have `self::$downloads->js` when the other mobile platforms don't?

mcdurdin: Because reasons. Will fix.

PR #63, opened 2019-12-09
feat: Update go links for desktop, developer, and linux

darcywong00:
While investigating to see if Keyman for Linux builds for the upcoming 20.04 LTS Focal Fossa, 
I was blocked on "Download" keyboard with the error:
> The resource you are looking for has been removed, had its name changed, or is temporarily unavailable.

This PR adds the go links needed for Keyman for Linux 13.0 development.

Per review comment, the go links for desktop, developer, and linux are each consolidated


go/linux/13.0/web.config
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
mcdurdin: Can we move this file up to the linux folder and use a match on `\d+.\d+` so we don't have to keep doing this for each version? If it diverges then we can make a change at that point...

We should do this for the Windows links as well

darcywong00: Done for Linux.

Question re: Windows links:
go/desktop/10.0/web.config has a redirect for issue-1285 that isn't in later versions. Should it be?

darcywong00: I went ahead and consolidated the desktop and developer go links 
(except for links to help.keyman.com/developer/13.0 which don't exist yet)

go/desktop/web.config
```diff
@@ -0,0 +1,67 @@
...
+<configuration>
+    <system.webServer>
+        <rewrite>
+            <rules>
+                <clear />
+
+                <!-- Links for Desktop 10.0 onward -->
+
+                <rule name="/go/desktop/X.0/download-keyboards" stopProcessing="true">
+                  <match url="^(10|11|12|13)\.0/download-keyboards" />
```
mcdurdin: Let's make the match on `([1-9][0-9])\.[0-9]` so we don't need to update this every release.

go/developer/web.config
```diff
@@ -0,0 +1,78 @@
...
+<configuration>
+    <system.webServer>
+        <rewrite>
+            <rules>
+                <clear />
+
+                <!-- Links for Developer 10.0 onward -->
+
+                <rule name="/go/developer/X.0/help-keyboards" stopProcessing="true">
+                  <match url="^(10|11|12|13)\.0/help-keyboards" />
```
mcdurdin: Let's make the match on `([1-9][0-9])\.[0-9]` so we don't need to update this every release.

go/developer/web.config
```diff
@@ -0,0 +1,78 @@
...
+                  <action type="Redirect" url="https://help.keyman.com/developer/keyboards/" />
+                </rule>
+                
+                <rule name="/go/developer/X.0/help-(mobile|packages)" stopProcessing="true">
+                  <match url="^(10|11|12)\.0/help-(mobile|packages)" />
```
mcdurdin: Similarly, we can assume the documentation will be available before release so let's include all versions here. We will always direct the help to the major version so the output redirect (hypothetical example 14.1 -> 14.0) is correct.

go/linux/web.config
```diff
@@ -5,29 +5,31 @@
             <rules>
                 <clear />
 
-                <rule name="/go/linux/12.0/download-keyboards" stopProcessing="true">
-                    <match url="^download-keyboards" />
+                <!-- Redirects for Keyman for Linux 11.0 onward -->
+
+                <rule name="/go/linux/X.0/download-keyboards" stopProcessing="true">
+                    <match url="^(11|12|13)\.0/download-keyboards" />
```
mcdurdin: Let's make the match on `([1-9][0-9])\.[0-9]` so we don't need to update this every release.

PR #65, opened 2019-12-16
Disable styling for "Rawang" in download links

### LINKED ISSUE #9: [keyman.com] Chrome squashing "Rawang" in language list footer
On the [main page](https://keyman.com) footer with many language names, Chrome is rendering "Rawang" as
image

even though it's Latin characters in an `raw` span.

This renders fine on Safari and Firefox
image

darcywong00:
Fixes #9

Applying the font family `MyanmarWeb` was causing "Krangku" to get squashed

With the fix:
image

### DIFF
--- a/cdn/dev/css/index.css
+++ b/cdn/dev/css/index.css
@@ -336,9 +336,11 @@
 	font-family: OriyaWeb;
 }
 /* Rawang */
+/*
 :lang(raw).lang-example {
 	font-family: MyanmarWeb;
 }
+*/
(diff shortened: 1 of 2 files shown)


cdn/dev/css/index.css
```diff
@@ -336,9 +336,11 @@
 	font-family: OriyaWeb;
 }
 /* Rawang */
+/*
```
darcywong00: I could have just deleted these, but I saw other languages were commented out

PR #66, opened 2019-12-17
feat: keep-in-touch on keyman.com

mcdurdin:
Moving keepintouch from tavultesoft.com to keyman.com.


desktop/keepintouch-100.php
```diff
@@ -0,0 +1,318 @@
...
+    </div>
+  </div>
+
+  <div id='mailchimp'>
+    <h2>Keyman Mailing List</h2>
+
+    <div class='container'>
+      <!-- Begin MailChimp Signup Form -->
+      <div id="mc_embed_signup">
+      <form action="https://tavultesoft.us1.list-manage.com/subscribe/post-json?u=99fcab2b035a8a51cd2158ca9&amp;id=7ccdac1e32&amp;c=?" method="get" id="mc-embedded-subscribe-form" name="mc-embedded-subscribe-form" class="validate"  novalidate>
```
darcywong00: Just double-checking the link is correct?

desktop/keepintouch-100.php
```diff
@@ -0,0 +1,318 @@
...
+    // This page is hosted in Keyman Desktop Configuration.
+    // Note that it must remain compatible with IE8/9 (yeuch).
+?>
+<!DOCTYPE HTML>
+<html>
+<head>
+  <meta charset='utf-8'>
+  <title>Keep in touch with the Keyman team</title>
+  <link href="https://fonts.googleapis.com/css?family=Cabin:400,400italic,500,600,700,700italic|Source+Sans+Pro:400,700,900,600,300|Noto+Serif:400" rel="stylesheet" type="text/css">
+  <script src="//code.jquery.com/jquery-1.11.2.min.js"></script>
```
darcywong00: thoughts on copying this to /cdn/dev/js/ ?

mcdurdin: We could do -- it would come as a later update rather than in this refactor. We will probably want to revisit this page for a future version of Keyman, too.

desktop/keepintouch-100.php
mcdurdin: Yes it is correct. Verified.

PR #79, opened 2020-02-18
feat(ios): adds 13.0 feature list, removes deprecated subpages

jahorton:
This should patch up the iOS content for a 13.0 release.

Note that the deleted subpages are _extremely_ outdated, likely having once served as the in-app help, which is now part of help.keyman.com.  Honestly, if something has somehow been liking to them, those links _should_ be broken.  (That said, PhpStorm didn't find anything.)


iphone-and-ipad/app/index.php
```diff
@@ -1,187 +0,0 @@
-<?php
```
darcywong00: fwiw, android/app/web.config has a redirect to help.keyman.com

PR #80, opened 2020-02-18
change: remove training registration link

darcywong00:
Registration is closed for the upcoming online Developer workshop, so updating the training page.


mcdurdin:
need to remove the link from the home page as well...

PR #82, opened 2020-02-18
Fix some prism styling and update go links for 13.0

darcywong00:
* Update 13.0 go links 
  * Involved refactoring mac links to redirect for all X.Y versions

* Fix styling on downloads page
* Update link to Keyman for Linux beta PPA


go/130/developer-help-mobile/index.php
```diff
@@ -0,0 +1,3 @@
+<?php
```
mcdurdin: Can we do redirects with web.config instead? I'm trying to move non-content like that into web.config as it's easier to maintain over time

darcywong00: Refactored to web.config.  

Also added 12.0 to the archive download links

go/web.config
```diff
@@ -10,6 +10,12 @@
                   <action type="Redirect" url="https://marc.durdin.net/2018/03/the-case-for-keyman/" />
                 </rule>
 
+                <!-- Developer 10.0 onward redirects for package guide -->
+                <rule name="developer-help-mobile / developer-help-packages" stopProcessing="true">
+                    <match url="^([1-9][0-9])([0-9])/developer-help-(mobile|packages)(/)?$" />
```
mcdurdin:
```suggestion
                    <match url="^(([1-9][0-9])([.]?)([0-9]))/developer-help-(mobile|packages)(/)?$" />
```

Suggest we make the [.] an optional addition so in later versions we can use a cleaner version number insertion. Also, if we ever did a 13.1 we need to use the whole version number

go/web.config
```diff
@@ -10,6 +10,12 @@
...
+                    <action type="Redirect" url="https://help.keyman.com/developer/{R:1}.0/guides/distribute/packages" />
```
mcdurdin:
```suggestion
                    <action type="Redirect" url="https://help.keyman.com/developer/{R:1}/guides/distribute/packages" />
```

This goes with that ... above

go/web.config
mcdurdin: I actually think later versions need to do `/go/developer/<version>/help-packages`

darcywong00: I can add your regex.

> I actually think later versions need to do `/go/developer/<version>/help-packages

go/developer/web.config already has
```
                <rule name="/go/developer/X.Y/help-(mobile|packages)" stopProcessing="true">
                  <match url="^([1-9][0-9])\.([0-9])/help-(mobile|packages)" />
                  <action type="Redirect" url="https://help.keyman.com/developer/{R:1}.0/guides/distribute/packages" />
                </rule>
```

mcdurdin:
Now wondering if the other ones are even used in latest Developer

Issue #41, opened 2019-10-01
Update privacy and terms of use policy links on all keyman.com sites


mcdurdin:
New [SIL International Language Software Privacy Policy](https://software.sil.org/language-software-privacy-policy) and [SIL International Language Software Terms of Use](https://software.sil.org/language-software-terms-of-use) documents posted.

Privacy Policy
- [x] keyman.com
- [x] help.keyman.com
- [x] keymanweb.com
- [x] Google Play Store listing

Terms of Use:
- [ ] ???

mcdurdin:
Marking this done.

PR #83, opened 2020-02-20
Update link to SIL privacy policy

darcywong00:
Addresses  #41  for updating link to SIL's privacy policy on keyman.com

### DIFF
--- a/privacy/index.php
+++ b/privacy/index.php
@@ -1,24 +1,3 @@
 <?php
-  require_once('includes/template.php');
-  
-  // Required
-  head([
-    'title' =>'Privacy Policy | Keyman',
-    'css' => ['template.css'],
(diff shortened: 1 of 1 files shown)


privacy/index.php
```diff
@@ -1,24 +1,3 @@
 <?php
-  require_once('includes/template.php');
-  
-  // Required
-  head([
-    'title' =>'Privacy Policy | Keyman',
-    'css' => ['template.css'],
-    'showMenu' => true
-  ]);           
+header('location: https://software.sil.org/language-software-privacy-policy/');
```
mcdurdin: Can we turn this into a web.config please? Trying to eliminate use of pure redirects in PHP for future maintenance.

PR #85, opened 2020-02-21
update linux installation instructions

darcywong00:
While I was validating the Keyman for Linux 13.0.103 release, I realized I needed to install ibus-keyman.
I would have expected it to be included in the keyman package...


ermshiperete:
That really should be a dependency on the package. The question is which package requires it?

PR #96, opened 2020-05-14
feat: Sentry integration

darcywong00:
Mirrors keymanapp/help.keyman.com#125 for this site

Adds reporting to sentry.keyman.com
Adds composer for integrating sentry


mcdurdin:
Replaced by #133 (almost the same, but uses common code now that we have it)

PR #101, opened 2020-05-29
fix(linux): Fix download on Linux from featured keyboards pages

ermshiperete:
There was still a piece missing to allow to download from Linux.

### DIFF
--- a/cdn/dev/js/kmlive.js
+++ b/cdn/dev/js/kmlive.js
@@ -135,7 +135,7 @@ function loaded(){
       } else {
         window.location.href = '/mac/download.php';
       }
-    }else if (platform == 'Windows') {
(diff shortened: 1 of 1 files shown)


cdn/dev/js/kmlive.js
```diff
@@ -135,7 +135,7 @@ function loaded(){
       } else {
         window.location.href = '/mac/download.php';
       }
-    }else if (platform == 'Windows') {
+    }else if (platform == 'Windows' || platform == 'Linux') {
       if (language && keyboard) {
```
darcywong00, line 139: Maybe Linux should get its own block like mac ?
```
} else {
  link = '/linux/download.php';
}
```

ermshiperete: done

PR #102, opened 2020-06-11
add more testimonials

darcywong00:
Gathered some testimonials from the Keyman Twitter feed.

We don't render Twitter emojis though

### DIFF
--- a/testimonials/index.php
+++ b/testimonials/index.php
@@ -11,13 +11,20 @@
 <h2 class="red underline">What people are saying about Keyman Desktop</h2>
 
 <p class="center">
-  <span class="testimonial">"@keyman, thank you for the Tamil keyboard. Loved how easy it is to type in Tamil.</span>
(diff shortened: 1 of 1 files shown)


testimonials/index.php
```diff
@@ -11,13 +11,20 @@
 <h2 class="red underline">What people are saying about Keyman Desktop</h2>
 
 <p class="center">
-  <span class="testimonial">"@keyman, thank you for the Tamil keyboard. Loved how easy it is to type in Tamil.</span>
+  <span class="testimonial">"I know most us had our first Amharic typing experience on #powergeez. It has been 10 years
```
mcdurdin, line 14: I'm not keen on referencing 'competitors'... I think it leads to sadness.

darcywong00: Edited to only keep the Keyman portion of the testimonial.

PR #103, opened 2020-06-17
feat: Reworked keyboard search

mcdurdin:
Depends on staging.api.keyman.com for local tests at present.

Changes in this PR:

* Rework various global hostname variables to use `$KeymanHosts` global. A future iteration may clean this up with a proper singleton class. (This also introduces the `Tier()` function for differentiation between dev/test/stage/prod.
* Strip out some old unused globals, e.g. `$robots`
* Load JS early because there were some race issues with the dynamic load
* Split search for legacy (Desktop 13.0 and earlier embedded search that relies on IE) and modern which strips out a lot of the guff needed for IE and simplifies the logic significantly. See _legacy folders and `go/desktop/web.config`.
* Fix a handful of broken links
* And yes, rework the keyboard search itself, marking results, adding download counters,


mcdurdin:
> I see that there are some newer files that seem to be duplicates of others, though placed underneath the `legacy` folder. Are there any changes I should be seeing, or is that all prep-work before making the changes?

The files under `legacy` are for version 1.0 of search (which is required for Keyman Desktop 13.0 and earlier embedded keyboard download).

mcdurdin:
I've updated the description of this PR to hopefully clarify things a bit. There are a few refactor things I could probably have split into a separate PR ... :(

PR #108, opened 2020-06-22
feat: highlight more search terms

mcdurdin:
Now supports highlighting any type of term supported by keyboard search.

### DIFF
--- a/cdn/dev/keyboard-search/search.js
+++ b/cdn/dev/keyboard-search/search.js
@@ -57,7 +57,7 @@ function wrapSearch(localCounter, updateHistory) {
     //hide_loading();
     $('#search-box').removeClass('searching');
     if(updateHistory && history.pushState) {
-      var r = q.match(/^(c(ountry)?|l(anguage)?)\:(iso|id)\:(.+)$/);
+      var r = q.match(/^(c|l)\:(id)\:(.+)$/);
       if(r && r[1].charAt(0) == 'c') {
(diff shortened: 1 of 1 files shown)


cdn/dev/keyboard-search/search.js
```diff
@@ -145,12 +149,15 @@ function process_response(q, res) {
...
-        case 'language': $('.title .match', k).text('('+kbd.match.name+' language)').mark(q); break;
-        case 'region': $('.title .match', k).text('('+kbd.match.name+')').mark(q); break;
-        case 'script': $('.title .match', k).text('('+kbd.match.name+' script)').mark(q); break;
-        case 'description': $('.description', k).mark(q); break;
+        case 'keyboard': $('.title a', k).mark(qq); break; // don't annotate
+        case 'keyboard_id': $('.id', k).mark(qq); break; // don't annotate
+        case 'language': $('.title .match', k).text('('+kbd.match.name+' language)').mark(qq); break;
+        case 'language_id': $('.title .match', k).text('(language tag \''+kbd.match.name+'\')').mark(qq); break;
+        case 'country': $('.title .match', k).text('('+kbd.match.name+')').mark(qq); break;
+        case 'country_id': $('.title .match', k).text('(country id \''+kbd.match.name+'\')').mark(qq); break;
```
ermshiperete, line 157: I guess it is intentional that `region` is missing now?

mcdurdin: Yes, we were using only countries so moving to that naming scheme. You'll see another PR later which does some additional renaming.

PR #111, opened 2020-06-22
feat: add bcp47 tag for embedded downloads

mcdurdin:
This will allow Keyman to automatically associate the appropriate language from the package at install time.

Part of solution for keymanapp/keyman#1456.

### DIFF
--- a/_includes/includes/ui/keyboard-details.php
+++ b/_includes/includes/ui/keyboard-details.php
@@ -66,6 +66,7 @@ define('GITHUB_ROOT', 'https://github.com/keymanapp/keyboards/tree/master/');
      * @param $id - keyboard ID
      * @param string $tier - ['stable', 'alpha', or 'beta']
      * @param bool $landingPage - when true, details won't display keyboard search box or title
+     * @param string $tag - BCP 47 tag to pass as a hint to download links for apps to make connection
      */
     public static function render_keyboard_details($id, $tier = 'stable', $landingPage = false, $tag = null) {
       self::$id = $id;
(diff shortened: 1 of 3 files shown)


_includes/includes/ui/keyboard-details.php
```diff
@@ -476,11 +481,11 @@ protected static function WriteDownloadBoxes() {
         }
       }
 
-      if ($webtext) {
-        echo "<h2 class='red underline'>Online tools</h2>" . $webtext;
-      }
-
       if ($embed == 'none') {
+        if ($webtext) {
```
mcdurdin, line 485: I opted to make the "use online" link visible only on the website, rather than also in the app, where it is a distraction.

cdn/dev/keyboard-search/search.js
```diff
@@ -145,7 +145,7 @@ function process_response(q, res) {
           "</div>"+
         "</div>");
 
-      $('.title a', k).text(kbd.name).attr('href', '/keyboards/'+kbd.id+embed_query_q);
+      $('.title a', k).text(kbd.name).attr('href', '/keyboards/'+kbd.id+(kbd.match.tag ? '?tag='+kbd.match.tag+embed_query_x : embed_query_q));
```
mcdurdin, line 148: embed_query_x starts with '&' and embed_query_q starts with '?'. This could be better but it's not worth rewriting for the sake of this!

_includes/includes/ui/keyboard-details.php
```diff
@@ -64,8 +67,9 @@ function downloadBinaryFile(a) {
      * @param string $tier - ['stable', 'alpha', or 'beta']
      * @param bool $landingPage - when true, details won't display keyboard search box or title
      */
```
darcywong00, line 69: Can you add a description for `@param string $tag`?

PR #114, opened 2020-06-23
fix: obsolete keyboards in search

mcdurdin:
For compatibility with pagination, obsolete keyboards are always shown but with red flag by them. The wording on the landing page has also been adjusted accordingly.

### DIFF
--- a/_includes/includes/ui/keyboard-details.php
+++ b/_includes/includes/ui/keyboard-details.php
@@ -309,7 +309,7 @@ END;
         $s = json_decode($s);
         if(is_object($s)) {
           self::$keyboard = $s;
-          self::$title = self::$keyboard->name;
+          self::$title = htmlentities(self::$keyboard->name);
           if (!preg_match('/keyboard$/i', self::$title)) self::$title .= ' keyboard';
           self::$description = isset(self::$keyboard->description) ? self::$keyboard->description : '';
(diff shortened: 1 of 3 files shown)


_includes/includes/ui/keyboard-details.php
```diff
@@ -457,12 +457,12 @@ protected static function WriteTitle() {
...
         $dep = self::$deprecatedBy;
-        $name = self::$title;
+        $id = self::$id;
         echo "
           <div class='download deprecated-new'>
             <a class='download-link' href='/keyboards/$dep$session_query_q'><span>Download</span></a>
-            <div class='download-title'>Download the latest version of $name</div>
-            <div class='download-description'>Click the Download button to get the latest version of this keyboard.</div>
+            <div class='download-title'>Keyboard '$id' is obsolete.</div>
+            <div class='download-description'>Click the Download button to get the replacement <span style='font-weight:bold'>$dep</span> keyboard instead.</div>
```
jahorton, line 465: So, for example... if someone finds the "Khmer (NiDA)" keyboard in search, would there still be a way to download that keyboard, or would attempting so only result in obtaining "Khmer Angkor" instead?

mcdurdin: This is what they are shown by default:

image

If they click the link `View details for old version of this keyboard` at the bottom there, they'll see:

image

(Or similar)

Note that this is not new functionality -- I've just tweaked the wording because I thought it was slightly unclear...

PR #116, opened 2020-06-24
chore: ground work for universal links, new autoload classes

mcdurdin:
This PR makes no substantial changes to the site.

What has changed:
* Adds a `/_includes/2020` folder for all autoloaded classes. These autoloaded classes should avoid use of `require()` (there are still a few little legacy dangly bits but I can clean those up over time).
* Tidies up a bunch of global variables/functions for future elimination (e.g. `Util::cdn()` vs `cdn()`, `KeymanVersion::stable_version` vs `$stable_version`. New pages must not use the old globals.
* Moves the `head`, `top_menu`, `phone_menu`, and `footer` templates into the `2020/templates` folder and makes them class-based.

New pages using this infrastructure should have a simple pattern and only ever use a *single* `require` statement (`ThisPage` should be named appropriately...). Some real examples will emerge with the universal links work in the next PR; but here's the concept:

```php
<?php
  declare(strict_types=1);

  namespace Keyman\Site\com\keyman;

  require __DIR__ . '/../_includes/autoload.php';
 
  use Keyman\Site\com\keyman\<class_foo>;

  ThisPage::render();

  class ThisPage {
    static function render() {
     ...
```

~~TODO before this is ready for review:~~

- [x] finish `Util::cdn()`
- [x] fixup `$stable_version`, `$beta_version` variables
- [x] add `PhoneMenu` template


cdn/dev/js/kmlive.js
```diff
@@ -357,4 +357,4 @@ function toggleDeprecatedVersionDetails() {
   $('#deprecated-old').toggle();
 }
 
-loaded();
+$(loaded);
```
mcdurdin, line 360: Need to wait until html is loaded before running `loaded()`; previously the script was delay-loaded at end of page but now it is loaded in the header.

jahorton:
It looks like this PR introduces more stringent server requirements than before.  One machine I tested with hosts `master` fine, but shows PHP errors when I checkout this branch:

> **Warning**: Unsupported declare 'strict_types' in **\keyman.com\_includes\2020\KeymanVersion.php** on line 2
>
> **Parse error**: syntax error, unexpected 'const' (T_CONST), expecting variable (T_VARIABLE) in **\keyman.com\_includes\2020\KeymanVersion.php** on line 8

Of course, there's a chance it's just my local (Windows) server config, but it _is_ a change, so I figured this may be worth noting.  My Mac setup (which was done more recently) loads this branch without issue.

mcdurdin:
> It looks like this PR introduces more stringent server requirements than before.

Good point. Yes, we need to move to PHP 7.4 for this change. We've already required this for the api.keyman.com changes.

PR #120, opened 2020-07-05
feat: keyboard install pages

mcdurdin:
# Universal Links - Landing Pages - Windows and Base Infrastructure

This PR includes: 

* Refactoring of basic template infrastructure to use modern PHP classes (2020/templates/ files)
* Rename of `tag` parameter to `bcp47` parameter for landing pages
* Refactoring of Keyboard Details pages classes for new layout
* Addition of new version of Bower (not IE compatible)
* Cleanup of ISAPI Rewrite rules to handle `/keyboard/install/<id>` routes
* Change over `/keyboards/<id>/share` route to `/keyboards/share/<id>`. This should not impact apps which should have been using the `/go/keyboard/<id>/share` route anyway.
* Removal of old code

## Pull Request Set

This is part of a set of five Pull Requests. The recommended order of review is:

* keymanapp/downloads.keyman.com#29 (basic infrastructure to support custom filenames for Keyman Windows Bootstrap)
* keymanapp/api.keyman.com#83 (APIs for the Windows bootstrap)
* keymanapp/keyman.com#120 (this PR -- universal link landing page + Windows)
* keymanapp/keyman#3304 (new Windows bootstrap)
* keymanapp/help.keyman.com#188 (documentation for Keyman Engine API changes)

# Keyboard Home Page - Old Version

image

# Keyboard Home Page - New Version

image

# Keyboard Install Page - Windows

image


_includes/2020/KeymanDownloadVersions.php
```diff
@@ -0,0 +1,40 @@
+<?php
```
mcdurdin, line 1: This class may not last long in this form; `KeymanDownloadVersions` is used currently *only* for macOS downloads and that will probably disappear as well once we finish the universal link model for macOS.

cdn/dev/keyboard-search/install.css
```diff
@@ -0,0 +1,15 @@
+/* List entries */
```
mcdurdin, line 1: This numbered list pattern is not really present, yet. It may be removed in future.

keyboards/install.php
```diff
@@ -0,0 +1,315 @@
...
+      self::$id = $id;
+      self::$bcp47 = self::validate_bcp47($bcp47);
+      self::$tier = self::validate_tier($tier);
+
+      self::LoadData();
+      self::WriteTitle();
+      if (isset(self::$downloads->kmp)) {
+        self::WriteDownloadBoxes();
+      } else {
+        //TODO: self::WriteNoDownloadMessage();
```
jahorton, line 59: Intentional omission?

_includes/async/keyboard-download.php
```diff
@@ -1,220 +0,0 @@
...
-  }
-
-  function fail($s) {
-    error_log('FAIL: '.$s);
-    exit(1);
-  }
-
-  $downloads = getKeyboardDownloadData($id);
-
-  recordGoogleAnalyticsEvent($cid, $id, $platform, $mode);
```
jahorton, line 33: I don't see where any Google analytics references are maintained under this PR's 'additions'; are we intentionally dropping use of Google Analytics for keyboard downloads?  Seems odd not to note that explicitly on the PR overview if so.

_includes/includes/ui/keyboard-details.php
```diff
@@ -103,121 +92,31 @@ protected static function map_license($s) {
...
-    protected static function download_box($url, $title, $description, $class, $linktitle, $platform, $mode='standalone') {
-      $urlbits = explode('/', $url);
-      $filename = array_pop($urlbits);
-      $id = self::$id;
-      global $embed;
-
-      $e_filename = urlencode($filename);
-      $e_id = urlencode($id);
-
-      if($embed != 'none') {
```
jahorton, line 115: We're no longer considering the `$embed` state for this downloading aspect?  If we are, where is the logic?

_includes/includes/ui/keyboard-details.php
```diff
@@ -478,44 +332,36 @@ protected static function WriteDownloadBoxes() {
...
-      $deviceboxfuncs = array(
-        "Windows" => "self::WriteWindowsBoxes",
-        "mac" => "self::WritemacOSBoxes",
-        "Linux" => "self::WriteLinuxBoxes",
-        "iPhone" => "self::WriteiPhoneBoxes",
-        "iPad" => "self::WriteiPadBoxes",
-        "Android" => "self::WriteAndroidBoxes"
-      );
+      // We'll write all the different platforms here and then let Bowser determine
+      // which box to show. This is true for both embedded and web-based viewing.
```
jahorton, line 336: I assume this is related to the previous comment (about where `$embed` is used regarding download boxes), but I'm still having trouble figuring out exactly where the connection point actually is.

_includes/async/keyboard-download.php
mcdurdin: That's a good catch. While the install.php page will capture keyboard downloads in Analytics, we would lose the events which we are going to be using for download statistics. I will add something into install.php (front-end to avoid missing stats through caching) to replicate this.

mcdurdin:
> I presume that despite the phrase "Universal Links" existing within the title, there's no actual implementation in this PR? Instead, it's work being undertaken to facilitate that goal?

Correct. `install.php` is the landing page for users when the app is not installed. In that sense, it's half of the Universal Links equation, plus the infrastructure to point to that page. As the title says, `Universal Links - Landing Pages - Windows and Base Infrastructure` -- it's really around the landing pages. (I could probably have dropped the `Windows` term.)

_includes/includes/ui/keyboard-details.php
mcdurdin: The embed and non-embed modes for downloading keyboards will be the same: we will request `/keyboards/install/<id>`, which the embedded dialog should capture and turn into a download request in the same way as the universal link handler. This does require changes to existing code.

Note that Windows, macOS and Linux 13.0 and earlier versions are already redirected to the legacy keyboard search so this should not impact them (they'll continue to use the old `keyman:` pattern).

keyboards/install.php
mcdurdin: The `TODO` indicates I need to do work here. It's on my list. Theoretically, a user should never get here if there is no .kmp as referring pages would not generate a link for a keyboard that has no .kmp (which would only be legacy web-only keyboards). Given that, it's low priority.

_includes/includes/ui/keyboard-details.php
mcdurdin: > connection point

I'm not sure what you mean?

mcdurdin:
Okay, I've restored analytics, as well as tidying up the end goal of having a single download point for packages that is platform independent and a safe, consistent URL.

The actual download URL (`/go/package/download/<id>`) is described in full at https://docs.google.com/document/d/1rhgMeJlCdXCi6ohPb_CuyZd0PZMoSzMqGpv1A8cMFHY/edit#heading=h.mio0p3wdzdye (Keyboard Package Downloads heading).

_includes/2020/templates/PlayStore.php
```diff
@@ -0,0 +1,31 @@
...
+
+    static function getTable() {
+      $url = self::url;
+      $img = self::getButtonImageSrc();
+      return <<<END
+<table class='app-store-links'>
+  <tbody>
+    <tr>
+      <td>
+        <a href="$url" target="itunes_store"><img id="app-store" src="{$img}" alt="Android app on Google Play"></a>
```
darcywong00, line 23: Can we use `target="play_store"`?

go/package/download.php
```diff
@@ -0,0 +1,152 @@
...
+*
+* Uses the best cryptographically secure method
+* for all supported pltforms with fallback to an older,
+* less secure version.
+* https://www.php.net/manual/en/function.com-create-guid.php#119168
+*
+* @param bool $trim
+* @return string
+*/
+function GUIDv4 ($trim = true)
```
darcywong00, line 120: I checked how [web-languageforge](https://github.com/sillsdev/web-languageforge) generates guids. They get https://github.com/ramsey/uuid from 
```
composer require ramsey/uuid
```

darcywong00: If you stick with GUIDv4, maybe split to a separate file?

_includes/includes/ui/keyboard-details.php
jahorton: How is `$embed` used by/forwarded to Bowser?

go/web.config
```diff
@@ -31,12 +32,17 @@
...
+                <rule name="download-package" stopProcessing="true">
+                  <match url="^package/download/([^/]+)$" />
+                  <action type="Rewrite" url="package/download.php?id={R:1}" appendQueryString="true" />
                 </rule>
 
                 <rule name="share-keyboard" stopProcessing="true">
                     <match url="^keyboard/([^/?]+)/share$" />
                     <!-- TODO: add a source platform for analytics? -->
-                    <action type="Redirect" url="/keyboards/{R:1}/share" />
+                    <action type="Redirect" url="/keyboards/share/{R:1}" />
```
darcywong00, line 45: Won't this redirect to a keyboard ID `share`?

go/package/download.php
```diff
@@ -0,0 +1,152 @@
...
+
+      if(empty($version)) {
+        // If version isn't provided, we'll query the live api for the version
+        $json = @file_get_contents(KeymanHosts::Instance()->api_keyman_com . '/keyboard/' . rawurlencode($id));
+        if ($json !== FALSE) {
+          $json = json_decode($json);
+        }
+
+        if(empty($json)) {
+          echo "Invalid parameter; id not found";
```
jahorton, line 39: Shouldn't this be more like the following?

```suggestion
          echo "Version parameter missing and could not be found automatically";
```

This `$json` corresponds to a version lookup, so why emit an error saying the _id_ couldn't be found here?

keyboards/install.php
```diff
@@ -0,0 +1,266 @@
...
+END;
+      return $result;
+    }
+
+    protected static function WriteMacOSBoxes() {
+      return self::download_box(
+        'Keyman for macOS',
+        KeymanDownloadVersions::getDownloadUrl('mac'), // note inconsistent platform name :(
+        htmlentities(self::$keyboard->name) . ' for macOS',
+        'Installs only ' . htmlentities(self::$keyboard->name) . '. <a href="/macosx">Keyman for Mac</a> must be installed first.',
```
darcywong00, line 131:
```suggestion
        'Installs only ' . htmlentities(self::$keyboard->name) . '. <a href="/mac">Keyman for macOS</a> must be installed first.',
```

go/web.config
darcywong00: nvm, I see the new redirects in web.config

darcywong00: Can you mention this change in the PR description?

_includes/includes/ui/keyboard-details.php
mcdurdin: `$embed` is not used by Bowser. Bowser sniffs the user-agent to determine the OS, browser, etc. We take the result from Bowser and set the `data-platform` attribute of the `<html>` element to activate the appropriate CSS rules.

_includes/2020/templates/PlayStore.php
mcdurdin: ok.

go/package/download.php
mcdurdin: For now, this is a fallback because some servers may not have `com_create_guid()`. I can move to a separate file but I will leave that for a future refactor I think (I am guessing that we'll always be using `com_create_guid()` on the live machines).

_includes/includes/ui/keyboard-details.php
jahorton: ... so users would see the same thing in iOS via in-app keyboard search as they would via web-surfing in Safari and searching outside the Keyman app?

keyboards/install.php
mcdurdin: The macOS part needs to be rewritten anyway -- that'll come as I do Universal Link work for mac. For now, only the Windows code is complete.

go/package/download.php
mcdurdin: If the api returns a fail, then the most likely reason is that the id we passed in was invalid. IMO other error conditions are so rare as to not worry about. 

It just so happens that we are only validating id when we do a version lookup, because invalid id will come out in the wash after the redirect anyway otherwise (but we need an exit route in this situation).

So the error message is reporting back to the page user a message that helps them understand that the parameters they passed were invalid. Perhaps it could be better worded, but the issue to the user is not around the version parameter, it's around the id parameter.

PR #123, opened 2020-07-09
feat: define go/windows paths

mcdurdin:
This goes with keymanapp/keyman#3325

### DIFF
--- a/go/desktop/web.config
+++ b/go/desktop/web.config
@@ -7,7 +7,7 @@
 
                 <!-- Links for Desktop 10.0 onward -->
 
-                <rule name="/go/desktop/X.Y/download-keyboards" stopProcessing="true">
+                <rule name="/go/desktop/13.0/download-keyboards" stopProcessing="true">
                   <match url="^(7|8|9|10|11|12|13)\.0/download-keyboards" />
                   <action type="Redirect" url="/_legacy/keyboards/?embed=windows" />
                 </rule>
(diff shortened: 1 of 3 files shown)


go/desktop/web.config
```diff
@@ -7,7 +7,7 @@
 
                 <!-- Links for Desktop 10.0 onward -->
```
darcywong00, line 8: To help prevent us from updating the wrong web.config file later, can you add a note that 14.0 onward updates should go in `windows/web.config`?

go/windows/web.config
```diff
@@ -0,0 +1,43 @@
...
+
+                <rule name="/go/windows/X.Y/issue-1285" stopProcessing="true">
+                  <match url="^([1-9][0-9]\.[0-9])/issue-1285" />
+                  <action type="Redirect" url="https://blog.keyman.com/2018/11/keyman-windows-10-1803-and-amharic-tigrinya-sinhala" />
+                </rule>
+
+                <!-- TODO: i18n needs to rewrite this to our home page for crowdin -->
+                <rule name="/go/windows/X.Y/create-locale" stopProcessing="true">
+                  <match url="^([1-9][0-9]\.[0-9])/create-locale" />
+                  <action type="Redirect" url="https://secure.tavultesoft.com/keyman/support/locale/" />
```
ermshiperete, line 28: Is it correct that this redirects to tavultesoft.com?

mcdurdin: Note the TODO comment for when we come to i18n. We have nowhere else ready to redirect to at present.

PR #130, opened 2020-07-22
feat: keyman protocol support in install page for Windows

mcdurdin:
Adds support for the `keyman:` protocol refresh pattern for Windows install/ processes.

TODO:

- [x] Disable this for IE and Legacy Edge because they kinda mess up

### DIFF
--- a/keyboards/install.php
+++ b/keyboards/install.php
@@ -90,35 +90,48 @@ END;
       $bcp47 = rawurlencode(empty(self::$bcp47) ? '' : self::BOOTSTRAP_SEPARATOR.self::$bcp47);
       $tier = self::$tier;
       $version = self::$versions->windows->$tier;
-      $e_keyboard_id = rawurlencode($keyboard->id);
-      $h_keyboard_name = htmlentities($keyboard->name);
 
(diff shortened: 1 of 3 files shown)


keyboards/install.php
```diff
@@ -90,35 +90,48 @@ protected static function WriteWindowsBoxes() {
...
+
+      $u = array_map('rawurlencode', $e);
+      $hu = array_map('htmlentities', $u);
+      $h = array_map('htmlentities', $e);
+
+      // Note, we don't need to capture an event for the keyboard download here, because we'll capture
+      // it when the bootstrap downloads the file.
+
+      // This maps to buildStandardWindowsDownloadUrl in install.js (which we don't use here for those
+      // few users who have JS disabled -- although this is not really a tested/supported scenario)
```
ermshiperete, line 111: Can this happen at all? If users don't have JS enabled this doesn't show because `keyboard-details.js` won't set the `data-platform` attribute.

mcdurdin: True... it doesn't hurt, though, and we can use a similar pattern later when we do the 'unknown platform' landing page (which is what appears if JS is disabled).

PR #135, opened 2020-07-23
feat: keyman protocol support in install page for Linux (keyman/#3271)

ermshiperete:


cdn/dev/keyboard-search/install.js
```diff
@@ -61,6 +61,46 @@ function startAfterPageLoad_Windows(data) {
...
+        (data.bcp47 == "" ? "" : "?bcp47=" + encodeURIComponent(data.bcp47));
+
+      const keymanUrl = buildStandardKeymanProtocolDownloadLink(
+        data.id, data.bcp47
+      );
+
+      location.href = keymanUrl;
+
+      const fallbackHandle = window.setTimeout(function() {
+        document.getElementById("content").innerHTML =
```
mcdurdin, line 93: I'd prefer to handle this with CSS controls so the HTML content is in the .PHP file, and we hide and show with CSS

ermshiperete: done

cdn/dev/keyboard-search/install.js
```diff
@@ -61,6 +61,34 @@ function startAfterPageLoad_Windows(data) {
...
+  }, 10);
+}
+
+/**
+ * Redirects to keyman:// to trigger Keyman Configuration, and if that fails,
+ * to the standard download url for the package to install. This currently only
+ * applies to Linux users.
+ * @param {object} data from the object with properties for host, tier, version, id, bcp47, name
+ */
+function startAfterPageLoad_Linux(data) {
```
mcdurdin, line 77: Given this is near identical to the Windows code above it, I think I'd like to see the two functions merged... I know it's a little more hassle but it will be less maintenance in the future :smile:

ermshiperete: Done in a separate PR.

mcdurdin: I am the very model of a consistent somethingorother.... Sorry!

PR #140, opened 2020-07-27
feat: server-side support for iOS universal links

jahorton:
Please note the documentation here: https://developer.apple.com/documentation/safariservices/supporting_associated_domains

Apple's file location _and_ name requirements are pretty stringent.  
- Yes, the file uses the JSON format.  
- No, it _must not_ use the .json extension.
- According to older documentation, we _may_ be able to drop the `.well-known` folder component of the URL.
- Any form of redirect leading to the file is not allowed.  The file itself must exist at the specified location, no redirects.

### DIFF
--- a/.well-known/apple-app-site-association
+++ b/.well-known/apple-app-site-association
@@ -1,0 +1,11 @@
+{
+  "applinks": {
+    "details": [
+      "appIDs": ["3YE4W86L3G.Tavultesoft.Keyman"],
+      "components": [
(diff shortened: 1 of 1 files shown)


jahorton:
Right, should have thought about web.config first.  Apple currently can't find the file... and on keyman.com.local, neither can I.  So there's probably a config setting that needs to happen.

mcdurdin:
See #141 for further spelunking and some baby steps

PR #141, opened 2020-07-28
feat: apple-app-site-association redux

mcdurdin:
This attempts to make the AASA file palatable. I am not convinced this is 100% correct.

* Branch IO AASA validation now passes: https://branch.io/resources/aasa-validator/
* Sadly, Apple's validator doesn't: https://search.developer.apple.com/appsearch-validation-tool. It gives the utterly useless message `"Error cannot parse app site association"`.

Changes I made:

1. I renamed the file and added an extension (only used internally, we rewrite the correct url)
2. I setup web.config to rewrite the .json file (this helps us with file management as MIME types are correct) to the /.well-known/apple-app-site-association url
3. I've fixed invalid JSON issues in the original file (missing object braces `{}` around `details`).
4. I've added both 9.0 and 13.0 format keys from the very muddled "documentation" (aka Stack Overflow and Forums):

   * "Documentation" regarding the two formats: https://developer.apple.com/forums/thread/129460
   * Ditto: https://stackoverflow.com/questions/32358098/the-apple-app-site-association-file-was-found-but-it-doesn-t-follow-the-required

Other wild links:
* Troubleshooting Universal Links: https://developer.apple.com/library/archive/qa/qa1916/_index.html#//apple_ref/doc/uid/DTS40017117-CH1-QATAG1
* iOS 9 format: https://developer.apple.com/library/archive/documentation/General/Conceptual/AppSearch/UniversalLinks.html#//apple_ref/doc/uid/TP40016308-CH12-SW1
* iOS 13 format: https://developer.apple.com/documentation/safariservices/supporting_associated_domains
* More detail on iOS 13: https://developer.apple.com/documentation/bundleresources/applinks/details
* "Documentation" regarding the two formats: https://developer.apple.com/forums/thread/129460
* Some more detailed 3rd party analysis: https://ios13.dev/universal-links-debugging-on-ios-13-cjwsux93w001p6ws1swtstmzc
* Another checklist: https://gist.github.com/andrewrohn/774185e4e15ddcc14f0a1e3c66c943e3

'Helpful' troubleshooting stories:
* https://developer.apple.com/forums/thread/120475
* https://developer.apple.com/forums/thread/123554

web.config Wilderness:
* https://docs.microsoft.com/en-us/iis/extensions/url-rewrite-module/url-rewrite-module-20-configuration-reference#setting-response-headers

Seems like the next step is to work through some of the troubleshooting (particularly related to swc) which I won't attempt tonight.

Note: I have manually uploaded the files in this PR to keyman-staging.com so you can test: https://keyman-staging.com/.well-known/apple-app-site-association or `curl -I https://keyman-staging.com/.well-known/apple-app-site-association` to see headers 😉


.well-known/web.config
```diff
@@ -0,0 +1,25 @@
...
+        <clear />
+        <rule name="apple-app-site-association">
+          <match url="^apple-app-site-association$" />
+          <action type="Rewrite" url="apple-app-site-association.json" />
+        </rule>
+      </rules>
+      <outboundRules>
+        <!-- We're going to force just application/json with no charset according to the Apple 'docs' -->
+        <rule name="Add mime type for apple-app-site-association">
+          <match serverVariable="RESPONSE_Content-Type" pattern='.' />
```
darcywong00, line 16: Does this impact the Android assetlinks.json that's now in the same folder?

jahorton:
... and it turns out the icons I used are iOS 13+ only.  *sigh*

Looking into alternatives now.

.well-known/web.config
mcdurdin: No, because of the condition on line 18.

.well-known/web.config
```diff
@@ -0,0 +1,25 @@
...
+          <conditions>
+            <add input="{REQUEST_URI}" pattern="apple-app-site-association" />>
```
mcdurdin, line 18:
```suggestion
            <add input="{REQUEST_URI}" pattern="apple-app-site-association" />
```

.well-known/apple-app-site-association.json
```diff
@@ -0,0 +1,19 @@
+{
+  "applinks": {
+    "apps": [],
+    "details": [
+      {
+        "appID": "3YE4W86L3G.Tavultesoft.Keyman",
+        "paths": ["/keyboards/install/*"],
+
```
jahorton, line 8:
```suggestion
}, {
```

jahorton: All examples I'm seeing online put the two definition styles in separate objects within the array, not within the same object.

We might should favor the top one and drop the bottom; the bottom is iOS 13.0+ friendly, whereas the top one supports older versions too.  Hopefully they can coexist, though.

.well-known/apple-app-site-association.json
```diff
@@ -0,0 +1,19 @@
+{
+  "applinks": {
+    "apps": [],
+    "details": [
+      {
+        "appID": "3YE4W86L3G.Tavultesoft.Keyman",
+        "paths": ["/keyboards/install/*"],
```
jahorton, line 7:
```suggestion
        "paths": ["/keyboards/install/*"]
```

jahorton: Sorry, forgot about that bit.  Also, I'm not so familiar with doing multi-line suggestions.

mcdurdin:
I've manually added those to the live site so we can test it. But the Apple validator still fails. Not clear if we need to wait for it to refresh its results or what.

jahorton:
> I've manually added those to the live site so we can test it. But the Apple validator still fails. Not clear if we need to wait for it to refresh its results or what.

I've seen people say it can take a couple hours for it to bother clearing its cache.

mcdurdin:
The validator has accepted the file as it stands now (note: keyman-staging.com has been updated so no longer has the test version of the files present, until we merge this PR)

PR #147, opened 2020-08-04
fix: skip deprecated keyboards

mcdurdin:
Added a link into the archive page to give access to obsolete keyboards. By default, they will not be listed.

Relates to keymanapp/api.keyman.com#87.

### DIFF
--- a/keyboards/index.php
+++ b/keyboards/index.php
@@ -66,6 +66,7 @@
   <form method='get' action='/keyboards' name='f'>
     <input id="search-q" type="text" placeholder="Enter language or keyboard" name="q" autofocus>
     <input id="search-f" type="image" src="<?= cdn('img/search-button.png"') ?>" value="Search" onclick="return do_search()">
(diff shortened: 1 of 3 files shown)


keyboards/index.php
```diff
@@ -66,6 +66,7 @@
   <form method='get' action='/keyboards' name='f'>
     <input id="search-q" type="text" placeholder="Enter language or keyboard" name="q" autofocus>
     <input id="search-f" type="image" src="<?= cdn('img/search-button.png"') ?>" value="Search" onclick="return do_search()">
+    <input id="search-obsolete" type="hidden" name="obsolete" value="0">
```
darcywong00, line 69: Does the keyboard search in /_includes/2020/templates/Menu.php also get updated (phone-menu and top-menu1)?

mcdurdin: No, this is the only interface where obsolete keyboard searches will be possible; if the field is missing, then it's treated as `0`

PR #149, opened 2020-08-05
fix: search history and hints

mcdurdin:
Relates to keymanapp/api.keyman.com#87.

* Search history is maintained more reliably now.
* Space is no longer trimmed while typing.
* Hints are shown when no search has been entered.

### DIFF
--- a/keyboards/index.php
+++ b/keyboards/index.php
@@ -80,7 +80,7 @@
   <ul>
     <li>The search always returns a list of keyboards. It searches for keyboard names and details, language names, country names and script names.</li>
     <li>You can apply prefixes <code>k:</code> (keyboards), <code>l:</code> (languages), <code>s:</code> (scripts, writing systems)
-    or <code>c:</code> (countries) to restrict your search results.</li>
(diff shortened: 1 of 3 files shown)


keyboards/index.php
```diff
@@ -71,6 +71,16 @@
...
+<div id='search-results-container'>
+<div id='search-results'></div>
+<div id='search-results-empty'>
+  <p>Enter the name of a keyboard or language to search for. (<a href="?q=p:*">Popular keyboards</a>)</p>
+  <br />
+  <p>Hints</p>
+  <ul>
+    <li>The search always returns a list of keyboards. It searches for keyboard names and details, language names, country names and script names.</li>
+    <li>You can apply prefixes <code>k:</code> (keyboards), <code>l:</code> (languages), <code>s:</code> (scripts, writing systems)
+    or <code>c:</code> (countries) to restrict your search results.</li>
```
darcywong00, line 83: Maybe (limit|narrow|filter) instead of restrict?

PR #162, opened 2020-09-18
feat: add model support to /go/package/download

### LINKED ISSUE #155: Update go/package/download to support models
Keyboard Package Downloads
Apps should download all packages from cloud using the following link pattern:

`https://keyman.com/go/package/download[/type]/package_id?[version=package_version][&platform=platform][&tier=tier][&bcp47=bcp47][&update=0|1]`

All parameters are optional. `platform`, `tier`, `bcp47` and `update` are used for analytics only.

* `package_version`: the version of the package to download. If not specified, the website will look up the latest version from downloads.keyman.com/api/keyboard and use that.
* `platform`: one of `android`, `ios`, `linux`, `macos`, `windows`.
* `tier`: one of `stable`, `beta`, `alpha`. If not specified, assumes `stable`. This should be retrieved from the version of Keyman that is installed.
* `bcp47`. A valid BCP 47 tag that is being associated with the keyboard. Typically used only for new installs.
* `update`. If `0`, reported as a new download (e.g. from a bootstrap installation). If `1`, then this is reporting upgrade of an existing package. If not present, defaults to `0`.
* **TODO**: `type`:  one of `keyboard`, `model`.  If not specified, assumes `keyboard`.

See also keymanapp/api.keyman.com#102 for related work

mcdurdin:
Fixes #155.

### DIFF
--- a/go/package/download.php
+++ b/go/package/download.php
@@ -32,12 +32,12 @@
         $type = 'keyboard';
       }
 
-      if(empty($id)) {
-        JsonApiFailure::InvalidParameters("id, version");
-      }
-
       if($type !== 'keyboard' && $type !== 'model') {
         JsonApiFailure::InvalidParameters("type");
+      }
+
+      if(empty($id)) {
+        JsonApiFailure::InvalidParameters("id, version");
       }
 
       if(empty($version)) {
--- a/go/package/download.php
+++ b/go/package/download.php
@@ -15,6 +15,7 @@
   KeymanComSentry::init();
 
   PackageDownloadPage::redirect_to_file(
+    isset($_REQUEST['type']) ? $_REQUEST['type'] : null,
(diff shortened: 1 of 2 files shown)


go/package/download.php
```diff
@@ -26,20 +27,28 @@
...
+    public static function redirect_to_file($type, $id, $version, $platform, $tier, $bcp47, $update, $ga_cookie) {
+      if(empty($type)) {
+        $type = 'keyboard';
+      }
+
       if(empty($id)) {
         JsonApiFailure::InvalidParameters("id, version");
       }
 
+      if($type !== 'keyboard' && $type !== 'model') {
```
darcywong00, line 39: Just wondering if this is intentional to validate $id (2nd parameter) before $type (1st parameter)?
Otherwise, I might move this to l. 34.

darcywong00 (approved):
lgtm. one minor question

PR #164, opened 2020-09-21
fix: menu disappears with mouse move

### LINKED ISSUE #45: [keyman.com] search for a keyboard on the home page
When searching for a keyboard, a dropdown menu may appear if you have search for some keyboards previously. The bug happens when the mouse pointer is moved to click on the desired keyboard--the menu disappears all of a sudden.

In the screenshot below, the user is trying to look for a language begin with "k" and the list appears, but when moving the pointer to "Khmer," the menu is gone leaving nothing to choose from.

image

mcdurdin:
Fixes #45. When the keyboard search within the Keyboards menu is focused, moving the mouse can cause the menu to disappear, which is confusing for the user.

### DIFF
--- a/cdn/dev/css/template.css
+++ b/cdn/dev/css/template.css
@@ -516,22 +516,28 @@ input[type="search"] {
 	overflow-y: visible;
 }
 
-#top-menu1 .menu-item:hover, #top-menu1 .menu-item img:hover{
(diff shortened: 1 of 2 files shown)


darcywong00 (approved):
lgtm

PR #166, opened 2020-09-21
fix: make obsolete banner visible at correct time

### LINKED ISSUE #163: Obsolete keyboards have banner to new keyboard hidden by default
The banner should not be hidden when the rest of the content is hidden. e.g. on page https://keyman-staging.com/keyboards/galaxiebiblescriptmnemonic

image

The banner shown here should be visible in the initial page load:

image

mcdurdin:
Fixes #163.


darcywong00 (approved):
lgtm

PR #172, opened 2020-09-21
fix: keyboard install page for macOS

mcdurdin:
Fixes the layout and links on the keyboard install page for macOS. Does not include any automatic install universal link
patterns. Also fixes wording for Linux page.

Part of #109.

image


darcywong00 (approved):
lgtm

PR #173, opened 2020-09-21
fix: minor bugs in search

mcdurdin:
Fixes keymanapp/api.keyman.com#107.

1. initial load of page does a redirect.
2. urls formed by getCurrentPath not as clean as they should be.
3. short text searches could be cancelled if enter pressed quickly.
4. minor refactoring and cleanup.


darcywong00 (approved):
lgtm

PR #177, opened 2020-09-22
chore: move landing pages to keyboards/h

### LINKED ISSUE #175: Dedicated landing pages don't really work well in embedded keyboard search
When working with the embedded search, links outside the `/keyboards` path are intended to open in an external browser. However, the dedicated landing pages are all currently hosted in the `/` path.

Propose moving the landing pages to `/keyboards/h/` as a future home for all keyboard landing pages. This would then involve updating `/web.config` to put in redirects for each of those pages. Organizationally, this reduces the level of confusion of the root of keyman.com, and puts dedicated landing pages on the same level as other keyboard redirects such as `/khmer`.

This also relates to #174.

mcdurdin:
Fixes #175. All keyboard landing pages are now under `/keyboards/h/` and the root paths redirect accordingly.

This should allow the embedded search dialogs to treat these links as internal to the search and not pop them into an external browser.


darcywong00 (approved):
lgtm

PR #178, opened 2020-09-22
fix: uasort instead of usort

mcdurdin:
I introduced a bug so I could do another 1 character PR. Really not really.


darcywong00 (approved):
gtm

PR #182, opened 2020-10-05
chore/collect php error.log in ci

### LINKED ISSUE #180: Collect back end PHP errors from php_errors.log during ci
We do this in the help.keyman.com ci, so copy this to the keyman.com ci

mcdurdin:
Fixes #180.

This copies the same solution as we had in help.keyman.com


darcywong00 (approved):
lgtm

PR #183, opened 2020-10-05
fix: linux - Clarify architecture

darcywong00:
https://community.software.sil.org/t/keyman-launch-pad-code-not-working/3920/3

A Linux user reported not being able to install Keyman for Linux on his configuration
> Im using Ubuntu Mate Release 20.04.1 LTS (Focal Fossa) 32 bit
Kernel Linux 5.4.0-1015-raspi armv7l
Mate 1.24.0

This clarifies Keyman for Linux's architecture as amd64.

TODO: copy to help.keyman.com/products/linux

### DIFF
--- a/linux/index.php
+++ b/linux/index.php
@@ -96,6 +96,18 @@
 
 <h2 class="red underline">Frequently Asked Questions</h2>
 <p>
+  <span class="red">Q.</span> What Linux distros will Keyman work with?
+</p>
+<p>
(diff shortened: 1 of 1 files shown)


mcdurdin (approved):
Do you think we should be building for arm7 as well?

darcywong00:
> Do you think we should be building for arm7 as well?

I've created keymanapp/keyman#3656 to explore that

PR #184, opened 2020-10-05
feat(ios): install page

mcdurdin:
Relates to keymanapp/api.keyman.com#109

iPhone 5 size:

image

iPad:

image

PR #185, opened 2020-10-05
feat(android): Android install page

mcdurdin:
Relates to keymanapp/api.keyman.com#109.

image

### DIFF
--- a/keyboards/install.php
+++ b/keyboards/install.php
@@ -53,36 +53,14 @@
 
       self::LoadData();
       self::WriteTitle();
-      self::WriteDownloadBoxes();
+
+      echo self::WriteWindowsBoxes();
+      echo self::WritemacOSBoxes();
+      echo self::WriteLinuxBoxes();
+      echo self::WriteAndroidBoxes();
+      echo self::WriteiOSBoxes();
+
       Foot::render();
-    }
-
-    protected static function download_box($keymanProductName, $keymanUrl, $title, $description, $class, $linktitle, $platform, $mode='standalone') {
-
-      if(isset(self::$keyboard->platformSupport->$platform) && self::$keyboard->platformSupport->$platform != 'none') {
-        $id = self::$id;
-
-        $e_id = urlencode($id);
-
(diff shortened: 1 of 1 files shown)


keyboards/install.php
```diff
@@ -136,7 +114,7 @@ protected static function WriteWindowsBoxes() {
       return $result;
     }
 
-    protected static function WriteMacOSBoxes() {
+    protected static function WritemacOSBoxes() {
```
darcywong00, line 117: Do you want to rename `legacy-keyboard-details::WriteMacOSBoxes` to be consistent?

keyboards/install.php
```diff
@@ -252,14 +230,56 @@ protected static function WriteLinuxBoxes() {
...
-        'android');
+      $keyboard = self::$keyboard;
+      $tier = self::$tier;
+      $version = self::$versions->android->$tier;
+
+      $e = [
+        'name' => $keyboard->name,
+        'host' => KeymanHosts::Instance()->downloads_keyman_com,
+        'tier' => $tier,
+        'version' => $version,
```
darcywong00, line 241:
```suggestion
        'version' => $keyboard->version,
```

Since `$version`  is the Keyman for Android version for the tier, I was getting a non-existent keyboard link
https://downloads.keyman.com/keyboards/sil_ipa/13.0.6218/sil_ipa.kmp

Do we still need `$version = self::$versions->android->$tier` if the "Install Keyman for Android" link goes straight to the Play Store?

darcywong00 (changes requested):
Is the keyboard version correct?

keyboards/install.php
mcdurdin: I won't touch the legacy code :)

keyboards/install.php
mcdurdin: Good pick! Fixed.

mcdurdin: Also fixed on iOS PR #184 and for Linux and macOS pages.

darcywong00 (approved):
lgtm

PR #187, opened 2020-10-08
chore: update keyman windows installer filenames

mcdurdin:
See also #186.

PR #188, opened 2020-10-08
chore: keyman windows installer filename (production)

mcdurdin:
This mirrors the work in #187, just for the production site. We only need to change the pre-release page for this update; stable versions of course haven't changed.

### DIFF
--- a/downloads/pre-release/index.php
+++ b/downloads/pre-release/index.php
@@ -40,7 +40,7 @@
 <p><a href='<?=$helphost?>/version-history'>Keyman version history</a> (all products)</p>
 
 <?php
-  downloadSection('Keyman Desktop for Windows', 'windows', 'keymandesktop-$version.exe', 'beta alpha');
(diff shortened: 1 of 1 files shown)


downloads/pre-release/index.php
```diff
@@ -40,7 +40,7 @@
 <p><a href='<?=$helphost?>/version-history'>Keyman version history</a> (all products)</p>
 
 <?php
-  downloadSection('Keyman Desktop for Windows', 'windows', 'keymandesktop-$version.exe', 'beta alpha');
+  downloadSection('Keyman Desktop for Windows', 'windows', 'keyman-$version.exe', 'beta alpha');
```
darcywong00, line 43: Until we get to 14.0 beta, won't the latest 13.0 beta be `keymandesktop-$version.exe`?

mcdurdin: Yeah but there is no current 13.0 beta so I don't think it matters.

darcywong00 (approved):
lgtm

PR #197, opened 2020-11-15
chore: fixup internal keyboard links

mcdurdin:
While the old links work, they involve an unnecessary redirect and make maintenance harder.


darcywong00 (approved):
lgtm

mcdurdin:
rebased to get broken link fix in and re-validate all links.

PR #203, opened 2020-11-27
fix: Chrome 85 breaks our clever download pattern

### LINKED ISSUE #200: bug: keyboard install page now downloads *and* triggers keyman: url
This seems to be a change in recent release of Chrome, perhaps?

https://keyman-staging.com/keyboards/install/khmer_angkor will show the keyman: url prompt but also start the download of the installer.

Firefox 81.0.2: works correctly.

mcdurdin:
Fixes #200.

The clever download pattern we established whereby the protocol handler would be used to start a download, and fallback to an installer if the protocol handler was not present, is broken with Chrome 85. It turns out that this was never explicitly supported, and was just a side-effect, so it's unlikely to return. This makes a marginally less smooth install experience but it should not be a significant issue, just one extra click on first install.

Screenshot of new install page, with Keyman already installed:

image

If the user does not have Keyman installed, they won't get the popup, and they'll need to click the Download button.

### DIFF
--- a/cdn/dev/keyboard-search/install.js
+++ b/cdn/dev/keyboard-search/install.js
@@ -57,6 +57,8 @@ function startAfterPageLoad_Windows(data) {
         location.href = keymanUrl;
       }
 
+      /*
+      Disabled, see https://github.com/keymanapp/keyman.com/issues/200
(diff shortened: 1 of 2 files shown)


darcywong00 (approved):
lgtm

PR #204, opened 2020-11-27
fix: array.prototype.find polyfill

mcdurdin:
Fixes keymanapp/keyman#3951.

Support old versions of Chrome on older Android devices in keyboard search.


darcywong00 (approved):
lgtm

Issue #210, opened 2020-11-30
bug: Bundled installer link is not working on keyman-staging.com when no bcp47 tag is specified


mcdurdin:
For example, the Install Keyboard link at https://keyman-staging.com/keyboards/install/sil_euro_latin?bcp47=fr&tier=alpha does currently work, but https://keyman-staging.com/keyboards/install/khmer_angkor?tier=alpha gives a broken pattern without a .exe suffix.

PR #214, opened 2020-12-02
fix: install link broken

mcdurdin:
The bundled installer link for Keyman keyboards for Windows was not working if bcp47 was not specified. Furthermore, it would default to the stable setup.exe, which is not correct while we are in alpha or beta for 14.0, as earlier stable versions do not support the custom filename pattern we need.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 **/.idea
+/.vscode
 cdn/deploy/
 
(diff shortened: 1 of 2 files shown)


jahorton (approved):

mcdurdin:
Belated: Fixes #210.

PR #219, opened 2020-12-04
feat(linux): Rename onboard package

ermshiperete:
Update instructions how to install pre-release version of keyman. The package name is now `onboard-keyman` instead of `onboard`. Part of keymanapp/keyman#3966.


mcdurdin (approved):
LGTM

PR #226, opened 2020-12-15
feat(linux): Fix name of beta ppa

ermshiperete:
Also update release notes (there are no packages for Fedora yet) and modernize installation command.

### DIFF
--- a/downloads/pre-release/index.php
+++ b/downloads/pre-release/index.php
@@ -48,8 +48,7 @@
 <p>Ubuntu: Keyman for Linux can be installed via <a href='https://launchpad.net/~keymanapp/+archive/ubuntu/keyman-daily'>launchpad</a>:</p>
 <pre class='language-bash code'><code>
 sudo add-apt-repository ppa:keymanapp/keyman-daily
-sudo apt-get update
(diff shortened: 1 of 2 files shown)


downloads/pre-release/index.php
```diff
@@ -48,8 +48,7 @@
 <p>Ubuntu: Keyman for Linux can be installed via <a href='https://launchpad.net/~keymanapp/+archive/ubuntu/keyman-daily'>launchpad</a>:</p>
 <pre class='language-bash code'><code>
 sudo add-apt-repository ppa:keymanapp/keyman-daily
-sudo apt-get update
-sudo apt-get install keyman onboard-keyman</code></pre>
+sudo apt install keyman onboard-keyman</code></pre>
```
darcywong00, line 51: How come the 14/index.php kept the `sudo apt upgrade` step?

darcywong00 (approved):
lgtm

downloads/pre-release/index.php
ermshiperete: I'm a little bit torn if that's necessary or not. Usually it shouldn't be necessary...

PR #227, opened 2020-12-22
feat: versioned download pages

mcdurdin:
This is a very basic landing page for versioned installs which we can use with keymanapp/keyman#4202. Depends on an update to downloads.keyman.com, coming shortly.

### DIFF
--- a/downloads/releases/_version_downloads.php
+++ b/downloads/releases/_version_downloads.php
@@ -1,0 +1,76 @@
+<?php
+  require_once('includes/template.php');
+  require_once('includes/ui/downloads.php');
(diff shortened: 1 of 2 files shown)


jahorton (commented):
So, this is for requesting a specific version of Keyman, rather than the latest version... right?

mcdurdin:
> So, this is for requesting a specific version of Keyman, rather than the latest version... right?

correct, /downloads/releases/a.b.c

mcdurdin:
Merging with Admin powers so that links aren't broken on the PR announcements

PR #231, opened 2021-01-13
chore: use php 7.4 in actions

mcdurdin:


.github/workflows/ci.yml
```diff
@@ -10,7 +10,16 @@ jobs:
       KEYMANHOSTS_TIER: TIER_TEST
 
     steps:
-    - uses: actions/checkout@v2
+    - name: Checkout
+      uses: actions/checkout@v2
+
+    - name: Setup PHP 7.4
+      uses: shivammathur/setup-php@v2
```
ermshiperete, line 17: I recently learned that from a security POV it would be better to use the git sha1 instead of the tag since tags in git can be changed.

ermshiperete (approved):

PR #233, opened 2021-01-18
fix: use full BCP 47 code for bookmarklet

mcdurdin:
Fixes keymanapp/api.keyman.com#46 and throws in a spinner on load.


darcywong00 (approved):
lgtm

PR #234, opened 2021-01-18
feat: bookmarklet search

mcdurdin:
Makes the bookmarklet list searchable because it is way long... oh, and tweaks the UI slightly, to make it slightly wider and less wasted real estate.

### DIFF
--- a/bookmarklet/index.php
+++ b/bookmarklet/index.php
@@ -23,26 +23,26 @@
 
 <h2 class="red underline">Keyman Bookmarklet</h2>
 <p>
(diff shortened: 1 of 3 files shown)


jahorton (commented):
Code inspection seems fine; looks like I need to update my local PHP before live-testing this, though.

mcdurdin:
> Code inspection seems fine; looks like I need to update my local PHP before live-testing this, though.

Or you could validate on keyman-staging.com if you prefer since it's a staging server :)

jahorton (approved):

PR #235, opened 2021-01-18
fix: bookmarklet list missing many keyboards

mcdurdin:
Fixes keymanapp/api.keyman.com#47.

This adds the version parameter to the api call in order to get all keyboards compatible with the current version. Also tidied up global variables for future maintainability.

There has long been a hidden feature which we have not really used, so I will highlight out again here: you can pass in a `keyboard` and `language` parameter into the page to get the page to show just a single keyboard, for example:

https://keyman.com/bookmarklet/?keyboard=tibetan_direct_input&language=adi-tibt

I can see this being useful in a variety of possible situations in the future.

### DIFF
--- a/bookmarklet/index.php
+++ b/bookmarklet/index.php
@@ -2,9 +2,10 @@
   require_once('includes/template.php');
   require_once __DIR__ . '/../_includes/autoload.php';
   use Keyman\Site\Common\KeymanHosts;
+  use Keyman\Site\com\keyman\KeymanVersion;
 
-  $lang = isset($_GET['language']) ? $_GET['language'] : '';
-  $kbd = isset($_GET['keyboard']) ? $_GET['keyboard'] : '';
+  $lang = isset($_GET['language']) ? json_encode($_GET['language']) : '""';
+  $kbd = isset($_GET['keyboard']) ? json_encode($_GET['keyboard']) : '""';
 
   // Required
   head([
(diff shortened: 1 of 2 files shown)


jahorton:
> There has long been a hidden feature which we have not really used, so I will highlight out again here: you can pass in a `keyboard` and `language` parameter into the page to get the page to show just a single keyboard, for example:
> 
> https://keyman.com/bookmarklet/?keyboard=tibetan_direct_input&language=adi-tibt
> 
> I can see this being useful in a variety of possible situations in the future.

Possible way to make it "less hidden":

Say I perform a keyboard search on keyman.com from my web browser, not in an embedded context.  For example, I might search for [Khmer keyboards](https://keyman.com/keyboards?q=khmer).  Then, on selecting one of the displayed keyboard options... [there's no mention of the bookmarklet there](https://keyman.com/keyboards/khmer_angkor).  The closest thing is a link to keymanweb.com using the keyboard, which _does_ offer this in the upper-right corner... _if_ it's visited from a desktop platform:

Screen Shot 2021-01-19 at 8 25 25 AM

If visited from a mobile platform, though:

Screen Shot 2021-01-19 at 8 26 50 AM

No mention or offering of bookmarklets there.

But here's the thing - especially with that "hidden feature" being a thing, why shouldn't the keyboard's page on keyman.com offer the bookmarklet if it's available for that keyboard?  In fact... might that be better than linking to keymanweb.com, assuming the page loads correctly and quickly?

jahorton (approved):

mcdurdin:
> But here's the thing - especially with that "hidden feature" being a thing, why shouldn't the keyboard's page on keyman.com offer the bookmarklet if it's available for that keyboard? In fact... might that be better than linking to keymanweb.com, assuming the page loads correctly and quickly?

Both options have value. keymanweb.com is a quick-and-easy place to use a keyboard. The bookmarklet is handy but doesn't work on all sites (e.g. FB now block 3rd party scripts from loading with their CSP). We'd need to reimplement the bookmarklet as a browser addin for better compatibility, which is certainly possible but outside the scope of this discussion.

I'm happy to consider further promotion of the bookmarklet feature for a future update of the website but equally I am happy for the hidden feature to remain hidden for the time being. We've got enough to do as it is 😉

mcdurdin:
> If visited from a mobile platform, though:

Bookmarklets on mobile are kinda hard to use.

PR #243, opened 2021-01-25
chore: Add release notes from B14S4 sprint retrospective

darcywong00:


mcdurdin (approved):
I'd also added the content to the Wiki but in a bit of a rush... This LGTM

PR #255, opened 2021-03-19
fix: Update deprecated links for featured keyboards

darcywong00:
Quite a few of the featured keyboards have since been deprecated.

This updates the keyboard ID's to get the current download links (along with working keymanweb links)

### DIFF
--- a/keyboards/h/sinhala/garp/index.php
+++ b/keyboards/h/sinhala/garp/index.php
@@ -49,7 +49,7 @@
     <a class="download-cta-more" href="/mac/">Learn more about Keyman for macOS</a>
   </div>
 </div>
-<div id="download-cta" data-language='si' data-keyboard='helabasauni'>
+<div id="download-cta" data-language='si' data-keyboard='helabasauni'> <!-- Is this basic_kbdsn1 ? -->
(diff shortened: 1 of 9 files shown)


keyboards/h/sinhala/garp/index.php
```diff
@@ -49,7 +49,7 @@
     <a class="download-cta-more" href="/mac/">Learn more about Keyman for macOS</a>
   </div>
 </div>
-<div id="download-cta" data-language='si' data-keyboard='helabasauni'>
+<div id="download-cta" data-language='si' data-keyboard='helabasauni'> <!-- Is this basic_kbdsn1 ? -->
```
mcdurdin, line 52: Not the same keyboards, no.  garpsinhala, helabasauni and basic_kbdsn1 are all unrelated -- apart from all being for Sinhala, that is.

mcdurdin (approved):
helabasauni may not be available any more, unsure.

keyboards/h/sinhala/garp/index.php
darcywong00: keymanweb.com doesn't load helabasauni :(

darcywong00: Just marked them as a TODO

PR #274, opened 2021-03-29
feat: zoom link for webinars

mcdurdin:
This adds a zoom link when it is time for the webinar, opening the link 5
minutes before the webinar starts. I think that makes sense -- or should we make it 15 minutes?

The link will disappear while webinars are not running, to avoid confusion.


darcywong00 (approved):
lgtm

PR #275, opened 2021-03-29
fix: use cdn versions of stylesheets etc in keyboard search

### LINKED ISSUE #272: bug: keyman.com/keyboards/x is showing all device download options on Android devices
The emulator shows only Android targets but loading on an actual device shows all targets, which is super confusing!

See https://community.software.sil.org/t/keyboard-package-error/4552/2

mcdurdin:
Fixes #272


darcywong00 (approved):
lgtm

PR #284, opened 2021-03-30
fix: support old versions of Firefox in install page

### LINKED ISSUE #276: bug: NS_ERROR_UNKNOWN_PROTOCOL: No error message
Sentry Issue: [KEYMAN-COM-3Z](https://sentry.keyman.com/organizations/keyman/issues/4235/?referrer=github_integration)

```
NS_ERROR_UNKNOWN_PROTOCOL: No error message
  at startAfterPageLoad_Linux/< (/cdn/deploy/keyboard-search/install.17ea5ad4b65082ad69c1a12060e16f37.js:89:7)
```

mcdurdin:
Fixes #276.

Ensures we don't try the protocol handler for old versions of Firefox, as older versions would try and load the page and break the install experience. Only version 87.0 and later will be supported for the protocol handler (current version as of writing).

### DIFF
--- a/cdn/dev/keyboard-search/install.js
+++ b/cdn/dev/keyboard-search/install.js
@@ -51,10 +51,24 @@ function startAfterPageLoad_Windows(data) {
         data.id, data.bcp47
       );
 
-      if(browser != "Internet Explorer" && browser != "Microsoft Edge Legacy") {
+      var firefoxVersion = 0;
+      if(browser == "Firefox") {
+        var match = /Firefox\/(\d+\.\d+)/.exec(navigator.userAgent);
+        firefoxVersion = match ? parseFloat(match[1]) : 0;
+      }
+
+      if(browser != "Internet Explorer" && browser != "Microsoft Edge Legacy" &&
+        (browser != "Firefox" || firefoxVersion > 86)) {
         // On IE and Edge Legacy, we will never try the keyman: protocol because
-        // it gives a poor user experience.
-        location.href = keymanUrl;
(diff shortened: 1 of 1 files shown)


cdn/dev/keyboard-search/install.js
```diff
@@ -71,6 +78,16 @@ function startAfterPageLoad_Windows(data) {
   }, 10);
 }
 
+function isLegacyFirefoxVersion() {
```
darcywong00, line 81: What about passing in `browser` as a parameter since it's already assigned in l.43?

mcdurdin: yeah but then I still need to do it in the Linux call so figured it made no real difference...

cdn/dev/keyboard-search/install.js
```diff
@@ -71,6 +78,16 @@ function startAfterPageLoad_Windows(data) {
...
+  const browser = document.documentElement.getAttribute('data-browser');
+  if(browser != "Firefox") return false;
+
+  var match = /Firefox\/(\d+\.\d+)/.exec(navigator.userAgent);
```
darcywong00, line 85: Could the regex determine the FF major version number, and then just compare with integer 87 in l.88?

mcdurdin: it could, but why?

darcywong00: I thought it would simplify the version comparison, but nvm

darcywong00 (approved):
lgtm

PR #296, opened 2021-04-26
fix: includes polyfill

mcdurdin:
Fixes [KEYMAN-COM-67](https://sentry.keyman.com/organizations/keyman/issues/4738) for various browsers.

### DIFF
--- a/cdn/dev/keyboard-search/search.js
+++ b/cdn/dev/keyboard-search/search.js
@@ -1,3 +1,19 @@
+// Polyfill for String.prototype.includes
+
+if (!String.prototype.includes) {
+  String.prototype.includes = function(search, start) {
+    'use strict';
+
+    if (search instanceof RegExp) {
+      throw TypeError('first argument must not be a RegExp');
+    }
+    if (start === undefined) { start = 0; }
+    return this.indexOf(search, start) !== -1;
+  };
+}
+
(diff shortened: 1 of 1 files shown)


cdn/dev/keyboard-search/search.js
```diff
@@ -1,3 +1,19 @@
+// Polyfill for String.prototype.includes
+
+if (!String.prototype.includes) {
+  String.prototype.includes = function(search, start) {
+    'use strict';
+
+    if (search instanceof RegExp) {
+      throw TypeError('first argument must not be a RegExp');
```
darcywong00, line 8: Is `throw new TypeError(...)` the same?

I saw that in a polyfill in cdn/dev/legacy-keyboard-search/search.js
https://github.com/keymanapp/keyman.com/blob/master/cdn/dev/legacy-keyboard-search/search.js#L19

mcdurdin: This code is copied straight out of [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes) and AFAICT, `throw Error('foo')` and `throw new Error('foo')` do the same thing (bleagh!).

image

darcywong00 (approved):
lgtm

PR #299, opened 2021-05-12
chore: rel=nofollow for package download links

mcdurdin:
This reduces the issue of web crawlers artificially bumping stats for keyboard package downloads.

Ref: https://ahrefs.com/blog/nofollow-links/


darcywong00 (approved):
lgtm

PR #300, opened 2021-05-17
chore: Add link to video of installing Keyman for macOS

darcywong00:
* Icon copied from webinar page


mcdurdin (commented):
Can we put the video link onto help.keyman.com instead please (that is, into the offline help)?

darcywong00:
Replaced by keymanapp/keyman#5108

PR #302, opened 2021-05-31
chore: https for dev hosts and secure session for keyboard search

mcdurdin:
Groundwork required for Chrome 90+ and SameSite cookies.

Fixes keymanapp/keyman#5193.

Without this, cookies within an iframe in the Download Keyboard dialog will be silently blocked, which causes styling and other aspects of the keyboard search to go wrong, as we use session variables to maintain the embedded state of the keyboard search, per platform.

In order to allow us to test locally, this means we need to upgrade our localhost (keyman.com.local, etc) sites to https. This is a good thing, overall, as it reduces the difference between local dev and live sites slightly further.

### DIFF
--- a/_common/KeymanHosts.php
+++ b/_common/KeymanHosts.php
@@ -88,7 +88,7 @@
         break;
       case KeymanHosts::TIER_DEVELOPMENT:
         $site_suffix = '.local';
-        $site_protocol = 'http://';
+        $site_protocol = 'https://';
         break;
       }
 
--- a/keyboards/session.php
+++ b/keyboards/session.php
@@ -1,5 +1,9 @@
 <?php
-  if(!isset($_SESSION)) session_start();
+  if(!isset($_SESSION)) {
+    session_set_cookie_params(["SameSite" => "None"]);   // Allow use in iframe, needed for Download Keyboards dialog
+    session_set_cookie_params(["Secure" => "true"]);     // None requires Secure to be set
+    session_start();
+  }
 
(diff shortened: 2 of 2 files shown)


rc-swag (approved):
lgtm

mcdurdin:
@MakaraSok, can you test that the keyboard search continues to work well on Keyman 14 (Windows at least, perhaps other platforms)?

MakaraSok:
The question is very broad to answer, but one thing that I notice when searching for a Khmer keyboard, i.e. Khmer (SIL) keyboard. One noticeable thing is that, only one button is shown on all platforms (i.e. Keyman for Windows, Keyman for macOS, Keyman for iOS and Keyman for Android, ) except web. Compare the screenshot below. 


On keyman.com, Chrome and Safari:
image

On Keyman 14.0.274, Windows 10 Pro 21H1:
image

On Keyman 14.0.274, macOS Big Sur:
image

On Keyman 14.0.274, Android 10:
image

On Keyman ???, iOS ???

mcdurdin:
> One noticeable thing is that, only one button is shown on all platforms (i.e. Keyman for Windows, Keyman for macOS, Keyman for iOS and Keyman for Android, ) except web

Yes, this is intentional. We don't want to pull users out of the flow of installing the keyboard when within the app. But from the website, they may well want to test the keyboard out without installing it.

MakaraSok:
OK. In that case, I don't notice anything problematic when searching for a keyboard on any of these platform. I haven't looked into Linux just yet, well in a long time actually.

PR #303, opened 2021-06-01
fix(windows): auto focus for keyboard search

mcdurdin:
Fixes keymanapp/keyman#5194.

Because the keyboard search lives inside an iframe when hosted in the Keyman app, we cannot use autofocus. Instead we use a script to do the autofocus at page load.

### DIFF
--- a/keyboards/index.php
+++ b/keyboards/index.php
@@ -32,6 +32,15 @@
 <script>
   var embed='<?=$embed?>';
   var embed_query='<?=$session_query?>';
+
+  if(embed != 'none') {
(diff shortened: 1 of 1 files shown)


rc-swag (approved):
lgtm

mcdurdin:
@MakaraSok, can you test that the keyboard search continues to work well on Keyman 14 (Windows at least, perhaps other platforms)?

MakaraSok:
Could you tell me what is the expectation of the fix and what is the issue really put in simple language for users so that I can effectively test. Thanks.

PR #304, opened 2021-06-06
feat(android): Google Play Install Referrer link

mcdurdin:
* Relates to keymanapp/keyman#5219.
* Goes with keymanapp/keyman#5230.

Add the link for Google Play Install Referrer and reorganise the install page for Android to make the one-step install clear.

Also cleans up a few bits and pieces with the look and feel of the install, fixing a HTML layout error and tweaking some CSS to make the page easier to understand.

## Before

image

## After

image

## Before - Keyboard Details page:

image

## After - Keyboard Details page:

image

Note that the keyboard search box at the top no longer wraps, and the left and right borders have been removed.

### DIFF
--- a/_includes/includes/ui/keyboard-details.php
+++ b/_includes/includes/ui/keyboard-details.php
@@ -301,8 +301,7 @@ END;
 
         <div id='search-box'>
           <form method='get' action='/keyboards' name='f'>
-            <div id='search-title'><a href='/keyboards'>Keyboard Search</a>:</div>
-            <input id="search-q" type="text" placeholder="(new search)" name="q">
+            <input id="search-q" type="text" placeholder="New keyboard search" name="q">
             <input id='search-page' type='hidden' name='page'>
             <input id="search-f" type="image" src="<?= cdn('img/search-button.png') ?>" value="Search">
           </form>
--- a/cdn/dev/css/template.css
+++ b/cdn/dev/css/template.css
@@ -2176,7 +2176,7 @@ button.feedback{
 
 @media all and (min-width: 10px) and (max-width: 600px){
 	.wrapper{
-		width: 320px;
+		width: auto;
 	}
 	/* Header */
 	#top-menu-bg{
--- a/keyboards/install.php
+++ b/keyboards/install.php
@@ -163,7 +163,7 @@ END;
       $downloadKeymanUrl = KeymanHosts::Instance()->keyman_com . '/mac/download';
 
       $result = <<<END
-        <div id='content' class='download download-macos'>
+        <div class='download download-macos'>
           <div>
             <p>If you have not yet installed Keyman for macOS, please install it first before installing the keyboard.</p>
             <ol>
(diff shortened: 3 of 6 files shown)


_includes/includes/ui/keyboard-details.php
```diff
@@ -301,8 +301,7 @@ protected static function WriteTitle() {
 
         <div id='search-box'>
           <form method='get' action='/keyboards' name='f'>
-            <div id='search-title'><a href='/keyboards'>Keyboard Search</a>:</div>
-            <input id="search-q" type="text" placeholder="(new search)" name="q">
+            <input id="search-q" type="text" placeholder="New keyboard search" name="q">
```
mcdurdin, line 304: Just to stop the search box wrapping in the keyboard details page

cdn/dev/css/template.css
```diff
@@ -2176,7 +2176,7 @@ button.feedback{
 
 @media all and (min-width: 10px) and (max-width: 600px){
 	.wrapper{
-		width: 320px;
+		width: auto;
```
mcdurdin, line 2179: Removes that left and right margin on small devices which is both ugly and unhelpful.

keyboards/install.php
```diff
@@ -227,12 +227,12 @@ protected static function WriteLinuxBoxes() {
                 <span>Install keyboard</span></a>
               </li>
             </ol>
-          </div>
```
mcdurdin, line 230: Oops. This extra `</div>` broke the remaining look and feel for the install page. Been this way for a little while now...

keyboards/install.php
```diff
@@ -215,7 +215,7 @@ protected static function WriteLinuxBoxes() {
       $downloadKeymanUrl = KeymanHosts::Instance()->keyman_com . '/linux/download';
 
       $result = <<<END
-        <div id='content' class='download download-linux'>
```
mcdurdin, line 218: The `id` value should be unique. Also, we weren't using it.

darcywong00:
So if the user already has Keyman installed, what happens when they click the green "Install from Play Store" for a keyboard?
(Is that from a keyboard search with the Chrome browser vs the in-app keyboard search?)

mcdurdin:
> So if the user already has Keyman installed, what happens when they click the green "Install from Play Store" for a keyboard?
> (Is that from a keyboard search with the Chrome browser vs the in-app keyboard search?)

Yes, this use-case is from a browser.

Normally, they won't get to this page if Keyman is already installed -- they'll be prompted to open the install page in the Keyman app. If they choose to open the install page in their browser instead, touching the Install Keyman + Keyboard link will open Google Play but this won't then install the keyboard: they'll have to click the second link instead.

It's possible that we might be able to check the Google Play Install Referrer again and get the new keyboard install details. I will test this.

darcywong00 (approved):
lgtm

mcdurdin:
I'm going to hold off merging this until 14.0.276 has been out for a week or so.

mcdurdin:
> I'm going to hold off merging this until 14.0.276 has been out for a week or so.

Or not ... tested and it is working well.

PR #319, opened 2022-01-27
fix: give iOS downloads for iOS Safari in desktop mode

mcdurdin:
When a user visits keyman.com in Safari on iPad or iPhone, they can request 'Desktop Mode'. In this situation, keyman.com was returning downloads for macOS instead of for iOS. This patch returns instead the iOS download if it detects a touch screen, which Macs do not support. If a touch-screen Mac is released, we'll have to revisit this.


rc-swag (approved):
lgtm

PR #343, opened 2022-09-19
feat: updated testimonials

mcdurdin:


darcywong00 (approved):
lgtm

mcdurdin:
I now don't think this video is appropriate for the home page. It isn't talking to Keyman users. It's talking to supporters / donors. It's also far too long for a home page video (which IMO should probably be no more than 1 minute and should engage the idea in the first 10 seconds). So returning to draft to rethink this.

mcdurdin:
Okay, I am going to refresh the testimonials but leave the video out for now. Perhaps we can link to it from the Get Involved page just so we have a public link?

PR #349, opened 2022-11-17
chore: basic security.txt

mcdurdin:
Have opted not to include a public key at this time, as that needs more administration to share with the responders to the email address.


darcywong00 (approved):
lgtm

PR #339, opened 2022-09-02
chore: migrate site from IIS to Apache 🦭

darcywong00:
Site migration of keyman.com for #337 to staging

Several TODO's before this base PR is ready:
* ~Keyboard redirects~
* ~Permanent redirects (301?)~ (#393)
* ~GHA link checker~ (#389)
* Incorporate [@keymanapp/shared-sites](https://github.com/keymanapp/shared-sites) bootstrapper

I noticed mdhost is in a different path compared to help.keyman.com, but we won't change that this round.

PR #356, opened 2023-01-19
fix: Add Docker for developer use 🦭

darcywong00:
This incorporates the Dockerfile and build.sh from help.keyman for developer use
Follows #339 

The steps are
```
build.sh build start
```

Then the site will be accessible at http://localhost:8053

So far, just the static content is working, as api.keyman needs a local instance for keyboard search...


darcywong00:
I'd like to get this into the staging-apache branch and continue to resolve redirects in a follow-on PR.

mcdurdin (approved):

PR #357, opened 2023-01-19
fix: override iPad's misreporting of user-agent in desktop mode

mcdurdin:
Relates to keymanapp/keyman#7915.


darcywong00 (approved):
lgtm

PR #371, opened 2023-06-02
chore: remove Google Analytics

mcdurdin:
GA requires an upgrade; we aren't really using the data so decided to just remove it instead.

See also https://github.com/keymanapp/keymanweb.com/pull/73


_common/KeymanHosts.php
```diff
@@ -107,12 +107,22 @@ function __construct() {
...
         $this->keymanweb_com = "https://keymanweb.com";
         $this->r_keymanweb_com = "https://r.keymanweb.com";
+      } else if($this->tier == KeymanHosts::TIER_TEST) {
+        $this->s_keyman_com = "https://s.keyman.com";
+        $this->api_keyman_com = "https://api.keyman.com";
+        $this->help_keyman_com = "https://help.keyman.com";
+        $this->downloads_keyman_com = "https://downloads.keyman.com";
+        $this->keyman_com = "https://keyman.com";
+        $this->keymanweb_com = "https://keymanweb.com";
+        $this->r_keymanweb_com = "https://r.keymanweb.com";
```
mcdurdin, line 117: GH actions were failing because staging api is not currently functional, so setup test to run with live backend api.keyman.com instead.

_common/KeymanHosts.php
```diff
@@ -107,12 +107,22 @@ function __construct() {
...
       } else {
         // TODO: allow override of these with e.g. KEYMANHOSTS_API_KEYMAN_COM='https://api.keyman.com';
         $this->s_keyman_com = "{$site_protocol}s.keyman.com{$site_suffix}";
         $this->api_keyman_com = "{$site_protocol}api.keyman.com{$site_suffix}";
+        // $this->api_keyman_com = "{$site_protocol}api.keyman.com"; //{$site_suffix}";
         $this->help_keyman_com = "{$site_protocol}help.keyman.com{$site_suffix}";
         $this->downloads_keyman_com = "{$site_protocol}downloads.keyman.com{$site_suffix}";
+        // $this->downloads_keyman_com = "{$site_protocol}downloads.keyman.com"; //{$site_suffix}";
```
mcdurdin, line 125: Left here because constantly fiddling with these currently -- once we move to docker this won't be needed.

jahorton (approved):
I presume we weren't using this to track monthly keyboard downloads or anything?  I figure that'd need some sort of analytics.  Just checking to ensure we don't break that.

mcdurdin:
> I presume we weren't using this to track monthly keyboard downloads or anything?

We used to use GA for this but switched to writing to our own db a long time ago, because the GA integration was fragile and overly complex.

PR #381, opened 2023-08-10
epic: Handle api.keyman.localhost and fix .htaccess 🦭

darcywong00:
Gradual improvements on getting the site fully Dockerized.

* Incorporate latest builder.inc.sh
* Incorporate trailing slash fixes from keymanapp/help.keyman.com#715
* Name the Docker container `keyman-com-app`
* Alter KeymanHosts.php and search.js to account for local api.keyman.com (http://localhost:8098 = http://host.docker.internal:8098) 
* Fix some keyboard searches

Note: CI link checker will still fail (haven't updated ci.yml to spool up the api.keyman.com Docker image, and the site still has keyboard links to resolve)


_common/KeymanHosts.php
```diff
@@ -119,14 +119,16 @@ function __construct() {
...
         $this->help_keyman_com = "{$site_protocol}help.keyman.com{$site_suffix}";
         $this->downloads_keyman_com = "{$site_protocol}downloads.keyman.com{$site_suffix}";
         // $this->downloads_keyman_com = "{$site_protocol}downloads.keyman.com"; //{$site_suffix}";
         $this->keyman_com = "{$site_protocol}keyman.com{$site_suffix}";
         $this->keymanweb_com = "{$site_protocol}keymanweb.com{$site_suffix}";
         $this->r_keymanweb_com = "https://r.keymanweb.com"; /// local dev domain is usually not available
       }
+      // temp overrides
+      $this->api_keyman_com = "http://localhost:8098";
+      $this->downloads_keyman_com = "https://downloads.keyman.com";
```
darcywong00, line 131: @mcdurdin - open to suggestions on how to handle the local "Docker" sites no longer named `.local`.

mcdurdin: I think we setup two things:

1. A docker image that reverse proxies to the various keyman.com domains from a single port.
2. Have an env var that specifies the port and protocol we want to use for that reverse proxy, defaulting to http + 8090; we set the env var to https + 443 on the live+staging environments.

Then each site can have its defined port, the reverse proxy has a record of which port is where, and we can use keyman-local.com as our local version:

* http://keyman-local.com:8090/
* http://s.keyman-local.com:8090/
* http://downloads.keyman-local.com:8090/
* http://web.keyman-local.com:8090/
* http://help.keyman-local.com:8090/
* http://api.keyman-local.com:8090/

On my machine, I may even decide to setup the reverse proxy to listen on :80, so I can skip the port altogether when using the local site.

We already own keyman-local.com.

darcywong00: I think these are now working with darcywong00/reverse-proxy.

I left some tweaks on KeymanHosts.php
local site protocol 'http://'
downloads uses the live https://downloads.keyman.com

mcdurdin: I wonder if we can move the reverse-proxy repo into keymanapp. Call it website-local-proxy perhaps? We'll all need it...

darcywong00: maybe it can go in keymanapp/resource/devbox/website-local-proxy/ ?

mcdurdin: no, I think a separate repo is fine, but in the keymanapp org.

.htaccess
```diff
@@ -114,7 +127,7 @@ RedirectMatch "^/(?!iphone-and-ipad)(ios|iphone|ipad)(\/.*)?" "/iphone-and-ipad$
 # Connect With Art landing page
 RedirectMatch "/connectwithart(\/|$)" "https://sites.google.com/sil.org/connectwithart/home"
 
-# per-language landing pages
+# per-language landing pages (maltese and rawang)
```
mcdurdin, line 130: Are the initial slashes appropriate?

.htaccess
```diff
@@ -133,12 +146,12 @@ RedirectMatch "/igbo(\/?)$"            "/keyboards/sil_nigeria_dot"
...
 RedirectMatch "/lao(\/?)$"             "/keyboards/basic_kbdlao"
 RedirectMatch "/malayalam(\/?)$"       "/keyboards/basic_kbdinmal"
-RedirectMatch "/maltese(\/?)$"         "/keyboards/maltese"
+RedirectMatch "^/maltese(\/?)$"         "/keyboards/maltese"
 RedirectMatch "/marathi(\/?)$"         "/keyboards/basic_kbdinmar"
 RedirectMatch "/mongolian(\/?)$"       "/keyboards/basic_kbdmon"
 RedirectMatch "/nepali(\/?)$"          "/keyboards/basic_kbdnepr"
 RedirectMatch "/oriya(\/?)$"           "/keyboards/basic_kbdinori"
-RedirectMatch "/rawang(\/?)$"          "/keyboards/rawang"
+RedirectMatch "^/rawang(\/?)$"          "/keyboards/rawang"
```
mcdurdin, line 154: This differs from all the others here. Afaict they should all start with ^ and not /

mcdurdin (approved):
Lgtm so far. Obviously we need to get tests passing to verify all the redirects

Reviewed on my phone

.htaccess
darcywong00: ok, standardized all the keyboard redirects to use `^/foo(/?)$`

.htaccess
darcywong00: The RedirectMatches wouldn't work with just `^foo`

Issue #337, opened 2022-08-06
feat: Transfer sites to Linux 🦭


darcywong00:
Per @mcdurdin 

> We have a bit of a priority on this now because PHP is ceasing to support Windows... We don't need to change PHP version, and I think nearly all the pages should be fine without changes. Primary things I can see are:
> * web.config -> .htaccess
> * Any places where we have backslashes in paths need to be forward slashes. Not likely to be many of these.
> * Consider downloadable file types, in case there are any issues there
>
>I don't think we have any executables in use on the site.
>Ubuntu version shouldn't be important, I hope -- just a reasonably recent one? It would be nice to be able to run them under Docker and/or Apache for Windows as well.

The sites that we need to transfer are:
- [x] https://help.keyman.com
- [ ] https://keyman.com
- [x] https://api.keyman.com (+ database)
- [x] https://keymanweb.com
- [x] https://s.keyman.com keymanapp/s.keyman.com#73
- [ ] Legacy sites:
  - [ ] https://blog.tavultesoft.com
  - [ ] https://keymankeyboards.com
  - [ ] https://r.keymanweb.com
  - [ ] https://tavultesoft.com

Reference
https://github.com/Azure/app-service-linux-docs/blob/master/Runtime_Support/php_support.md

Some [engineering documentation](https://docs.google.com/document/d/16hUSAO6zdKqHTiEqJXk-h3889a87CnmT6uJZwQ7FF1I/edit#) for @mcdurdin and @tim-eves

darcywong00:
Jotting a list of TODO's for the keyman.com site:
* Apply permanent redirects
* keyboard redirects
* /.well-known/web.config

mcdurdin:
IIS rewrite notes. Edit \windows\system32\inetsrv\config\applicationHost.config, add following section under `<system.webServer>` (edit `<rewrite>` if it already exists):

```xml
        <rewrite>
            <globalRules>
                <rule name="help.keyman-local.com forward proxy" stopProcessing="true">
                    <match url=".*" />
                    <conditions>
                        <add input="{HTTP_HOST}" pattern="^help\.keyman-local\.com|help\.keyman\.com\.local$" />
                    </conditions>
                    <action type="Rewrite" url="http://localhost:8055/{R:0}" />
                </rule>
                <rule name="s.keyman.com proxy" stopProcessing="true">
                    <match url=".*" />
                    <conditions>
[...]

PR #397, opened 2023-10-18
chore: Generate CDN in init-container.sh 🦭

darcywong00:
Continuing #337

* Fix cdn pathing for Docker container
* Generate CDN in init-container.sh

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -38,16 +38,8 @@ function test_docker_container() {
   echo "Done checking links"
 }
 
-# Custom cleanup of CDN
-function clean_docker_container_cdn() {
-  clean_docker_container $KEYMAN_IMAGE_NAME $KEYMAN_CONTAINER_NAME
-
-  # Cleanup CDN
-  rm -rf cdn/deploy
-}
-
 builder_run_action configure  bootstrap_configure
-builder_run_action clean      clean_docker_container_cdn
+builder_run_action clean      clean_docker_container $KEYMAN_IMAGE_NAME $KEYMAN_CONTAINER_NAME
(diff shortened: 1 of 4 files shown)


build.sh
```diff
@@ -38,8 +38,16 @@ function test_docker_container() {
   echo "Done checking links"
 }
 
+# Custom cleanup of CDN
+function clean_docker_container_cdn() {
+  clean_docker_container $KEYMAN_IMAGE_NAME $KEYMAN_CONTAINER_NAME
+
+  # Cleanup CDN
+  rm -rf cdn/deploy
```
mcdurdin, line 46: We probably don't need to clean this up -- it's only a cache so doesn't break things if it is present.

cdn/cdnrefresh.php
```diff
@@ -36,7 +36,7 @@ function generate_hashed_version_file($source, $deploy, $source_file) {
       $dest_file = $source_file_short;
     }
     if(!file_exists($deploy . $dest_file)) {
-      echo $source_file . " => " . $deploy . $dest_file . "\n";
+      echo $source_file . " => " . $deploy . $dest_file . "<br/>";
```
mcdurdin, line 39: This is a CLI process, why emit html?

mcdurdin (approved):
LGTM apart from those two nits :grin:

cdn/cdnrefresh.php
darcywong00: I used https://github.com/keymanapp/keymanweb.com/blob/0dba6472d3300ef15a4820ee53953d0051333aa0/cdn/cdnrefresh.php#L39 as reference

darcywong00: Reverted to "\n"

mcdurdin: gotcha

PR #399, opened 2023-10-25
chore: Add test page for legacy keyboard searches

darcywong00:
This primarily adds a test page for testing `/_legacy/keyboards` links. 

Additional updates
* Use showMenu header flag for Markdown files
* Add `[QSA]` (query string append) flags to correspond to web.config `appendQueryString="true"`
* Reorganize the /test pages to go, legacy, keyboard (keyboard links to be checked/updated on follow-on PR. For now, it's mostly a copy of legacy links)


_legacy/.htaccess
```diff
@@ -1,31 +1,31 @@
...
 
-# "legacy /keyboards/{name} to /keyboards/keyboard.php"
-RedirectMatch "/_legacy/keyboards/(.*)" "/keyboards/keyboard.php?id=$1"
+# "legacy /keyboards/{name} to keyboards/keyboard.php
+RewriteRule "^keyboards/(.*)" "/_legacy/keyboards/keyboard.php?id=$1" [END]
 
-# "legacy /keyboard/{name} to /keyboards/..." stopProcessing="true">
-RedirectMatch "/_legacy/keyboard/(.*)$" "/keyboards/$1"
+# "legacy /keyboard/{name} to keyboards/...
+RedirectMatch "/_legacy/keyboard/(.*)$" "/_legacy/keyboards/$1"
```
darcywong00, line 31: These didn't seem to match the original comments, but it's what I had to do to get the links to match what's on the live site (currently using https://github.com/keymanapp/keyman.com/blob/master/_legacy/web.config)

.htaccess
```diff
@@ -110,13 +110,13 @@ RewriteRule "^keyboards/(?!index\.php|install|keyboard|session|share)([^/]+)$" "
 RewriteRule "^keyboards/languages/(.*)" "/keyboards/index.php?q=l:id:$1" [L]
 
 # /keyboards/download to /keyboards/download.php
-RewriteRule "^keyboards/download(.php)?" "/keyboards/download.php" [L]
+RewriteRule "^keyboards/download(.php)?" "/keyboards/download.php" [L,QSA]
```
mcdurdin, line 113: `L,QSA` and not `END,QSA`?

mcdurdin (approved):

.htaccess
darcywong00: It probably is [END,QSA]. I'll change in follow-on

PR #400, opened 2023-10-25
fix: keyboard links

darcywong00:
Follows #399 

This exercises more of the keyboard redirects on /test/keyboards


test/keyboards.md
```diff
@@ -24,75 +19,72 @@ showmenu: false
...
+[live: keyboards/download.php?id=sil_ipa](https://keyman.com/keyboards/download.php?id=sil_ipa)
 
-### legacy /keyboards/id
+### keyboards/legacy
 
-[_legacy/keyboards/sil_ipa](/_legacy/keyboards/sil_ipa)
+[keyboards/legacy/cameroon querty non-u kbd-fonts](/keyboards/legacy/cameroon%20qwerty%20non-u%20kbd-fonts)
 
-[live: _legacy/keyboards/sil_ipa](https://keyman.com/_legacy/keyboards/sil_ipa)
+[live: keyboards/legacy/cameroon querty non-u kbd-fonts](https://keyman.com/keyboards/legacy/cameroon%20qwerty%20non-u%20kbd-fonts)
```
darcywong00, line 84: I'm not sure what legacy keyboard ID would exercise these links.

.htaccess
```diff
@@ -61,20 +61,21 @@ RewriteRule "^downloads/releases(\/)?$" "/downloads" [R,END]
...
 
 # /keyboards/{id}/ to /keyboards/id
-RedirectMatch 301 "^keyboards/([^/]+)/$" "/keyboards/$1"
+# Remove final /
+RewriteRule "^keyboards/([^/]+)/$" "/keyboards/$1" [R=301,END,QSA]
 
-# /keyboards/ to /keyboards
-RedirectMatch 301 "^keyboards/$" "/keyboards"
+# /keyboards/ to /keyboards (Handled by other redirect)
+# Remove final /
```
darcywong00, line 72: I was getting infinite redirects if I left this in

.htaccess
```diff
@@ -61,20 +61,21 @@ RewriteRule "^downloads/releases(\/)?$" "/downloads" [R,END]
...
 
 # Old share url /keyboards/{id}/share[/] to /keyboards/share/id
-RedirectMatch "^keyboards/(?!install|download|share)/share(/?)$" "/keyboards/share/$1"
+RewriteRule "^keyboards/(.*)/share(/?)$" "/keyboards/share.php?id=$1" [R=301]
```
darcywong00, line 75: Instead of redirecting to a redirect, I short-circuit to

https://github.com/keymanapp/keyman.com/blob/adc4ede646cbad7bb866f1a36349016cfadc6b61/web.config#L120-L124

mcdurdin: Yeah, but the original is redirecting to a rewrite (not to a redirect), so it was:

1. `keyboards/khmer_angkor/share` redirects to `keyboards/share/khmer_angkor`
2. `keyboards/share/khmer_angkor` which rewrites to `/keyboards/share.php?id=khmer_angkor`

We don't want `/keyboards/share.php?id=khmer_angkor` to be visible to the visitor, just `/keyboards/share/khmer_angkor`. So I think we need to keep the original pattern?

test/keyboards.md
mcdurdin: legacy could use modern keyboard ids -- it's about old versions of Keyman

darcywong00: Hmm - still unclear what should happen. The live site
https://keyman.com/keyboards/legacy/sil_ipa

seems blank too.

mcdurdin (changes requested):
Let's update the Broken Link Checker to start on /test/index.md, and ensure /test/index.md has links to /test/go.md, /test/keyboards.md, and / (root). That way we can exercise these links also.

mcdurdin:
Also, can we consider calling /test into /_test so that it is clearly for internal use?

darcywong00:
Link checker is flagging 
http://localhost:8053/go/developer/16.0/language-lookup
which is moved to ethnlogue.com

mcdurdin:
> Link checker is flagging
> http://localhost:8053/go/developer/16.0/language-lookup
> which is moved to ethnlogue.com

Okay, let's skip this link, because the redirect is not followed, which is why BLC is returning BLC_UNKNOWN -- it gets the base link but then doesn't follow the redirect so it doesn't know if it is valid or not.

_test/go.md
```diff
@@ -201,12 +201,13 @@ Direct help to the major version
 [live: keyman-engine-home](https://keyman.com/go/developer/16.0/keyman-engine-home)
 
 
-### /go/developer/X.Y/language-lookup
-
-[language-lookup](/go/developer/16.0/language-lookup)
-
-[live: language-lookup](https://keyman.com/go/developer/16.0/language-lookup)
+### Skipping /go/developer/X.Y/language-lookup
+ethnlogue has a redirect that the link checker does not follow. BLC is returning BLC_UNKNOWN -- it gets the base link but then doesn't follow the redirect so it doesn't know if it is valid or not.
```
mcdurdin, line 205:
```suggestion
keyman.com has a redirect to ethnologue that the link checker does not follow. BLC is returning BLC_UNKNOWN -- it gets the base link but then doesn't follow the redirect so it doesn't know if it is valid or not.
```

mcdurdin (approved):
LGTM

PR #402, opened 2023-11-13
fix: Use getenv() instead of _SERVER for vars

darcywong00:
Follows the pattern of keymanapp/api.keyman.com#213
in replacing $_SERVER vars with $env.

I left default _SERVER vars (and HTTP_ vars) intact
https://www.php.net/manual/en/reserved.variables.server.php

### DIFF
--- a/_includes/includes/template.php
+++ b/_includes/includes/template.php
@@ -6,6 +6,8 @@
   require_once(__DIR__ . '/../2020/KeymanVersion.php');
   require_once(__DIR__ . '/../2020/templates/Head.php');
 
+  $env = getenv();
+
   function template_finish($foot) {
     //ob_end_flush();
 
@@ -16,8 +18,9 @@
 
   function head($args=[]){
     // Args are title='My Page Title', css='page.css' showMenu=true/false, showHeader=true/false, foot=true/false
+    global $env;
 
-    $agent = isset($_SERVER['HTTP_USER_AGENT']) ? $_SERVER['HTTP_USER_AGENT'] : '';
+    $agent = isset($env['HTTP_USER_AGENT']) ? $env['HTTP_USER_AGENT'] : '';
     // Get device
     if (strstr($agent,'Windows')) {
         $device = 'Windows';
--- a/_includes/includes/template.php
+++ b/_includes/includes/template.php
@@ -6,8 +6,6 @@
   require_once(__DIR__ . '/../2020/KeymanVersion.php');
   require_once(__DIR__ . '/../2020/templates/Head.php');
 
-  $env = getenv();
-
   function template_finish($foot) {
     //ob_end_flush();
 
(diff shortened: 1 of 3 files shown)


_includes/includes/template.php
```diff
@@ -16,8 +18,9 @@ function template_finish($foot) {
 
   function head($args=[]){
     // Args are title='My Page Title', css='page.css' showMenu=true/false, showHeader=true/false, foot=true/false
+    global $env;
 
-    $agent = isset($_SERVER['HTTP_USER_AGENT']) ? $_SERVER['HTTP_USER_AGENT'] : '';
+    $agent = isset($env['HTTP_USER_AGENT']) ? $env['HTTP_USER_AGENT'] : '';
```
mcdurdin, line 23: Per https://www.php.net/manual/en/reserved.variables.server.php:

> In addition to the elements listed below, PHP will create additional elements with values from request headers. These entries will be named `HTTP_` followed by the header name, capitalized and with underscores instead of hyphens. For example, the Accept-Language header would be available as `$_SERVER['HTTP_ACCEPT_LANGUAGE']`.

```suggestion
    $agent = isset($_SERVER['HTTP_USER_AGENT']) ? $_SERVER['HTTP_USER_AGENT'] : '';
```

So this entire file can be rolled back

_includes/includes/template.php
```diff
@@ -16,8 +18,9 @@ function template_finish($foot) {
 
   function head($args=[]){
     // Args are title='My Page Title', css='page.css' showMenu=true/false, showHeader=true/false, foot=true/false
+    global $env;
```
mcdurdin, line 21:
```suggestion
```

_includes/includes/template.php
```diff
@@ -6,6 +6,8 @@
   require_once(__DIR__ . '/../2020/KeymanVersion.php');
   require_once(__DIR__ . '/../2020/templates/Head.php');
 
+  $env = getenv();
```
mcdurdin, line 9:
```suggestion
```

keyboards/h/sinhala/index.php
```diff
@@ -1,5 +1,6 @@
 <?php
-    $agent = isset($_SERVER['HTTP_USER_AGENT']) ? $_SERVER['HTTP_USER_AGENT'] : '';
+    $env = getenv();
+    $agent = isset($env['HTTP_USER_AGENT']) ? $env['HTTP_USER_AGENT'] : '';
```
mcdurdin, line 3: This is in $_SERVER, so roll back this file :grin:

mcdurdin (changes requested):

mcdurdin (approved):

mcdurdin:
One thing about global `$env` variable is that if it is set in multiple files, then any changes to it will be overwritten. `$_SERVER` is never reset. I saw a pattern in one PR where a value in `$env` was replaced, but that could cause us grief...

darcywong00:
> One thing about global `$env` variable is that if it is set in multiple files, then any changes to it will be overwritten. `$_SERVER` is never reset. I saw a pattern in one PR where a value in `$env` was replaced, but that could cause us grief...

ah, from keymanapp/api.keyman.com#213
```
      $env = getenv();
      $force = !empty($env['TEST_REBUILD']);
      unset($env['TEST_REBUILD']); // wow, but means this only gets run once :)
```

Issue #403, opened 2023-11-15
chore: Docker migration TODO 🦭


mcdurdin:
Outstanding work that needs to be completed for the Docker-based/k8s-based website host transfer:

## help.keyman.com
- [x] http -> https upgrade on staging and production -- verify no needless .htaccess rules ➡ keymanapp/help.keyman.com#876
- [x] verify TIER on k8s ➡ keymanapp/help.keyman.com#877
- [x] staging: Use fleet to automatically deploy 
- [x] staging: Use pinned Docker images ➡keymanapp/help.keyman.com#890
- [x] remove legacy resources/kubectl folder (help.keyman.com repo) ➡ keymanapp/help.keyman.com#883
- [x] api/deploy
  - [x] rename api/deploy to _control/webhook/deploy (site-kubernetes-config repo) ➡ keymanapp/site-kubernetes-config#2
  - [x] serialized git pull -- preventing parallel git pulls/deploys
  - [x] don't forget post-pull tasks such as
    - [x] generating kbd index in init-container.sh
    - [x] CDN refresh in init-container.sh ➡ keymanapp/help.keyman.com#911
  - [x] update webhook endpoint on github
  - [x] Verify that rolling updates incl. timeouts etc is working well with keyman-staging
- [ ] production: Deploy
  - [x] Use fleet to automatically deploy (in site-kubernetes-config / rancher) @tim-eves 
  - [x] point production cluster to staging branch initially (in site-kubernetes-config) @tim-eves 
  - [x] Test locally @tim-eves 
  - [x] RAM limits @tim-eves 
  - [x] Turn off Azure deployment
  - [x] ensure *.keyman.com, keymanweb.com is in the CF origin cert loaded onto the production cluster
  - [x] Add production deploy webhook endpoint on github
  - [x] Merge staging -> master (in help.keyman.com repo)
  - [x] Swap production cluster to master branch (in site-kubernetes-config repo)
  - [x] Test locally
  - [x] webhook on github (production)
  - [x] Update DNS for help.keyman.com (on Cloudflare)
  - [ ] Remove from Azure

Post-Deployment:
- [ ] staging: for backend, use api-keyman-com.keyman to bypass Cloudflare
- [ ] production: for backend, use api-keyman-com.keyman to bypass Cloudflare

## keyman.com
- [x] /.well-known/web.config #404
  - [x] remove let's encrypt bits
  - [x] test remaining rules in .htaccess
- [x] verify TIER on k8s ➡ #405
[...]

PR #412, opened 2024-01-26
chore: Add _control/ready

darcywong00:
Addresses the point for https://github.com/keymanapp/keyman.com/issues/403

>keyman.com
>
> * test web server ready, and _common files, vendor files ready
> * .htaccess rewrite from /_control/ready to /_control/ready.php


mcdurdin (approved):
LGTM

We should consider a unit test for these at some point too

PR #414, opened 2024-02-01
fix: go links for model/keyboards 🦭

darcywong00:
Oops - I built a local Keyman Android app to try to download keyboard packages from keyman-staging and realized I had a few TODO's to handle in /go links

Reference to the missing bits from web.config (currently used on the live site)
https://github.com/keymanapp/keyman.com/blob/ab3d0a4dfcb1915a9dafd10e2d98fff0a90b4d03/go/web.config#L19-L47

* Also updated the _test/go.md test page

### DIFF
--- a/go/.htaccess
+++ b/go/.htaccess
@@ -5,14 +5,27 @@ RedirectMatch "/go/why\/?$" "https://marc.durdin.net/2018/03/the-case-for-keyman
 # Developer 10.0 onward redirects for package guide
 RedirectMatch "/go/(([1-9][0-9])([.]?)([0-9]))/developer-help-(mobile|packages)(/)?" "https://help.keyman.com/developer/$1/guides/distribute/packages"
 
-# TODO: Download redirects for keyboard permalinks (these three rules need refresh)
+# Download redirects for keyboard permalinks (TODO: these three rules need refresh)
+
+# download-kmp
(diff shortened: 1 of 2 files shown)


go/.htaccess
```diff
@@ -5,14 +5,27 @@ RedirectMatch "/go/why\/?$" "https://marc.durdin.net/2018/03/the-case-for-keyman
 # Developer 10.0 onward redirects for package guide
 RedirectMatch "/go/(([1-9][0-9])([.]?)([0-9]))/developer-help-(mobile|packages)(/)?" "https://help.keyman.com/developer/$1/guides/distribute/packages"
 
-# TODO: Download redirects for keyboard permalinks (these three rules need refresh)
+# Download redirects for keyboard permalinks (TODO: these three rules need refresh)
```
mcdurdin, line 8: What's the TODO about?

mcdurdin (approved):

go/.htaccess
darcywong00: I don't know what the TODO for the 3 rules involve. They were carried over from the web.config.

https://github.com/keymanapp/keyman.com/blob/ab3d0a4dfcb1915a9dafd10e2d98fff0a90b4d03/go/web.config#L21-L35

mcdurdin: Lost in the mists of time. Probably unimportant, so we won't worry about it at this time.

PR #424, opened 2024-02-13
chore: Add info to .htaccess

darcywong00:


mcdurdin:
This isn't needed, https://keyman.com/_control/info already works due to the global .php extension stripping. In fact, we don't need the rule for ready.php either!

.htaccess
```diff
@@ -44,7 +44,8 @@ RewriteRule "^downloads/releases/(?!_version_downloads\b)(.+)$" "/downloads/rele
 # index
 RewriteRule "^downloads/releases(\/)?$" "/downloads" [R,END]
 
-# Ready
+# Control rules
+RewriteRule "^_control/info$" "_control/info.php" [END]
 RewriteRule "^_control/ready$" "_control/ready.php" [END]
```
mcdurdin, line 49:
```suggestion
```

mcdurdin (approved):
Suggest removing the control rules, then LGTM for the additions to info.php

PR #429, opened 2024-02-19
fix: Redirected queries missing parameters

mcdurdin:
Fixes keymanapp/keyman#10746.

`RedirectMatch` does not forward parameters; instead need to use `RewriteRule`.

### DIFF
--- a/go/android/.htaccess
+++ b/go/android/.htaccess
@@ -2,7 +2,7 @@
 # Links for Android 14.0 onward
 
 # /go/android/X.Y/download-keyboards/languages"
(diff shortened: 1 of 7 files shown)


darcywong00 (approved):
lgtm (assuming the link checker passes)

mcdurdin:
> lgtm (assuming the link checker passes)

link checker doing its job!

```
Getting links from: http://localhost:8053/_test/go
├─BROKEN─ http://localhost:8053/go/windows/16.0/archived-downloads (HTTP_404)
```

PR #430, opened 2024-02-19
fix: append query strings for some other redirects

darcywong00:
Follows #429 
Audited the .htaccess Redirect rules to see if I missed other instances where query string needs to be appended.

### /.htaccess
https://github.com/keymanapp/keyman.com/blob/ab3d0a4dfcb1915a9dafd10e2d98fff0a90b4d03/web.config#L39-L47

https://github.com/keymanapp/keyman.com/blob/ab3d0a4dfcb1915a9dafd10e2d98fff0a90b4d03/web.config#L59-L62

### /go/download/.htaccess

https://github.com/keymanapp/keyman.com/blob/ab3d0a4dfcb1915a9dafd10e2d98fff0a90b4d03/go/download/web.config#L8-L11

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -20,10 +20,10 @@ AddType application/json apple-app-site-association
 RewriteRule "^.well-known/apple-app-site-association$" "/.well-known/apple-app-site-association.json" [L]
 
 # macosx and macos to mac (ignore case)
-RedirectMatch 301 "^(?i)/(macosx|macos)\b(.*)$" "/mac$2"
+RewriteRule "^(macosx|macos)\b(.*)$" "/mac$2" [NC,R=301,END,QSA]
 
 # Redirect deprecated Google Plus link
-RedirectMatch 301 "^(?i)/plus.*" "/"
+RewriteRule "^plus.*" "/" [NC,R=301,END,QSA]
 
 # /donate -> donate.keyman.com
 RedirectMatch 301 "^(?i)/donate(\/.*)?" "https://donate.keyman.com"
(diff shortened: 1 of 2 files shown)


.htaccess
```diff
@@ -20,10 +20,10 @@ AddType application/json apple-app-site-association
 RewriteRule "^.well-known/apple-app-site-association$" "/.well-known/apple-app-site-association.json" [L]
 
 # macosx and macos to mac (ignore case)
-RedirectMatch 301 "^(?i)/(macosx|macos)\b(.*)$" "/mac$2"
+RewriteRule "^(macosx|macos)\b(.*)$" "/mac$2" [NC,R=301,END,QSA]
 
 # Redirect deprecated Google Plus link
-RedirectMatch 301 "^(?i)/plus.*" "/"
+RewriteRule "^plus.*" "/" [NC,R=301,END,QSA]
```
mcdurdin, line 26: We could just delete this, Google+ is so dead.

.htaccess
```diff
@@ -20,10 +20,10 @@ AddType application/json apple-app-site-association
 RewriteRule "^.well-known/apple-app-site-association$" "/.well-known/apple-app-site-association.json" [L]
 
 # macosx and macos to mac (ignore case)
-RedirectMatch 301 "^(?i)/(macosx|macos)\b(.*)$" "/mac$2"
+RewriteRule "^(macosx|macos)\b(.*)$" "/mac$2" [NC,R=301,END,QSA]
```
darcywong00, line 23: For case-insensitive matches, RedirectMatch uses `(?i) and RewriteRule uses `NC`

mcdurdin (approved):

.htaccess
mcdurdin: .htaccess really is an appalling file format, isn't it? So many edge cases and fiddly bits. So much legacy. Seems like a lot of the functionality "just happened" rather than being properly designed.

PR #448, opened 2024-04-26
chore: correct <li> styling and make bookmarklet search box responsive

Meng-Heng:
Please review. Thank you!

### DIFF
--- a/bookmarklet/index.php
+++ b/bookmarklet/index.php
@@ -59,12 +59,14 @@
     </p>
 </div>
 
+<div>
 <h2 class="red underline">Use the bookmarklet on any website</h2>
-<ol>
-    <li>Load any web page, such as a search engine</li>
-    <li>Click the Keyman Bookmarklet in your Bookmarks toolbar or menu</li>
-    <li>You may need to wait a second or two for Keyman Bookmarklet to load. Then click in the text field you wish to type into. The Keyman Bookmarklet user interface will appear and you can type in your language</li>
-</ol>
+  <ol  id="bookmarklet-use">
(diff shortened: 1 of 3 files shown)


mcdurdin:
Could you include screenshots of before and after style? Thanks!

Meng-Heng:
Sure!
This is the before:
image
This is the after:
image
Thank you!

bookmarklet/index.php
```diff
@@ -59,12 +59,14 @@
     </p>
 </div>
 
+<div>
```
darcywong00, line 62: I see the previous sections declare the id on the `<div>` instead of the inner contents (lines 42, 50).

Should we use `id="bookmarklet-use"` here?

darcywong00 (approved):
lgtm

Just a nit question about where id's go

Meng-Heng:
Darcy, can you test run to see if the style has been applied? No matter what I tried on my laptop, the new style didn't apply. I had to use the terminal for guidance 😄.

This is how it should look like:

- I switch the media style to the template.css
- Add more responsiveness to the bookmarklet-search
- index.css has only the bookmarklet-use ol styling

screencapture-keyman-localhost-8053-bookmarklet-2024-04-30-12_16_52

Meng-Heng:
Darcy, this should look good now. Should I press the merge PR?

darcywong00:
oh right. the sizing looked good to me. I couldn't find where the styling index.css `#bookmarklet-use o` got used though

Meng-Heng:
It is being used here:
image

PR #467, opened 2024-08-30
chore: minor tweaks to all-versions content

mcdurdin:
Rename all-version to all-versions, plus some very minor text tweaks.


darcywong00 (approved):
lgtm

PR #471, opened 2024-09-10
fix: Skip CDN for --debug builds

### LINKED ISSUE #470: chore: build.sh --debug should not build the CDN
When building for local development, we want to be able to use the --debug flag so that the CDN is not used and we read directly from /cdn/dev instead of /cdn/deploy for resources such as images and stylesheets.

This means that build.sh start should not build the CDN at all, because if the file is missing, it'll be read from /cdn/dev as a fallback.

darcywong00:
Fixes #470 to skip generating CDN if `./build.sh --debug`

Since this is changing `BOOTSTRAP_VERSION` to `chore/v0.10`, I'm unsure if we want to merge this to master. Or have a separate branch while the Logo work is in flux...

(edit)
BOOSTRAP_VERSION v0.10 is published so updating PR to use that.

### DIFF
--- a/resources/init-container.sh
+++ b/resources/init-container.sh
@@ -1,7 +1,11 @@
 #!/usr/bin/env bash
 
-echo "---- Generating CDN ---"
-rm -rf cdn/deploy
-cd cdn
-php -d include_path=/var/www/html/_includes:. cdnrefresh.php
-cd ..
+if [[ ! $1 =~ "debug" ]]; then
+  echo "---- Generating CDN ---"
+  rm -rf cdn/deploy
+  cd cdn
+  php -d include_path=/var/www/html/_includes:. cdnrefresh.php
+  cd ..
+else
+  echo "Skip Generating CDN"
+fi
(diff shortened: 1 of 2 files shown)


mcdurdin (approved):
I think hold off merging until v0.10 is finished

darcywong00:
> I think hold off merging until v0.10 is finished

This will wait on keymanapp/shared-sites#28

resources/init-container.sh
```diff
@@ -1,7 +1,11 @@
...
-cd ..
+if [[ ! $1 =~ "debug" ]]; then
+  echo "---- Generating CDN ---"
+  rm -rf cdn/deploy
+  cd cdn
+  php -d include_path=/var/www/html/_includes:. cdnrefresh.php
+  cd ..
+else
+  echo "Skip Generating CDN"
+fi
```
darcywong00, line 11: The keymanapp/help.keyman.com#1563 version also includes this
```suggestion
  rm -rf ../cdn/deploy
fi
```

PR #475, opened 2024-09-25
fix: Modify stable url for keyboard download to kmp file

### LINKED ISSUE #474: bug: Stable url for keyboard download no longer works
Our sldr files use this form of url to reference a keyboard, but it just goes to the home page for the keyboard:
https://keyman.com/go/keyboard/sil_euro_latin/download/kmp

This one works: https://keyman.com/go/package/download/sil_euro_latin

However, @mcdurdin says the first one is recommended and the second one is not a publicly stable one. 

This is possibly a bug coming out of our move to Docker with Apache recently.

darcywong00:
Fixes #474 

From the reported issue, 
> Our sldr files use this form of url to reference a keyboard, but it just goes to the home page for the keyboard:
> https://keyman.com/go/keyboard/sil_euro_latin/download/kmp
>
> This one works: https://keyman.com/go/package/download/sil_euro_latin

We want to change /go/keyboard/**/download/kmp to match the behavior  of /go/package instead of going to a keyboard landing page.

This modifies the rule in .htaccess to match what /go/package does.
https://github.com/keymanapp/keyman.com/blob/b695ba05961de47b25ba38ba3a210a232b09f615/go/.htaccess#L27-L28

The links can be manually compared on
https://github.com/keymanapp/keyman.com/blob/b695ba05961de47b25ba38ba3a210a232b09f615/_test/go.md?plain=1#L348-L354

Until this PR is merged, localhost will redirect to the kmp file while the live site still goes to the keyboard page.


go/.htaccess
```diff
@@ -8,7 +8,8 @@ RedirectMatch "/go/(([1-9][0-9])([.]?)([0-9]))/developer-help-(mobile|packages)(
 # Download redirects for keyboard permalinks (TODO: these three rules need refresh)
 
 # download-kmp
-RedirectMatch "/go/keyboard/([^/?]+)/download/kmp$" "/keyboards/download?id=$1&platform=windows&mode=standalone"
+# Modified to match download-package below
+RewriteRule "^keyboard/([^/]+)/download/kmp$" "package/download.php?type=keyboard&id=$1" [END,QSA]
 
 #download-exe
 RedirectMatch "/go/keyboard/([^/?]+)/download/exe$" "/keyboards/download?id=$1&platform=windows&mode=bundle"
```
mcdurdin, line 15: What about this rule and download-js? They also seem to be wrong?

mcdurdin (changes requested):
Hmm, is this a much older bug, not actually coming out of the Apache/docker transition? Let's make sure we get it right for all patterns. The bundled installer .exe is now actually not a "bundle" but a file name pattern which causes the bundle to download (see https://kmn.sh/kb35, keyman-setup filename patterns section), so we may need to do some more work to get that right.

go/.htaccess
mcdurdin: For .exe, look at the pattern generated e.g. by https://keyman.com/keyboards/install/khmer_angkor, and use that: https://downloads.keyman.com/windows/stable/17.0.331/keyman-setup.khmer_angkor.exe

mcdurdin: download-js should go to downloads.keyman.com/keyboards/name/version/name.js, e.g. https://downloads.keyman.com/keyboards/thamizha_bamini/2.2/thamizha_bamini.js

darcywong00: These will be split into #495

mcdurdin (approved):

Issue #488, opened 2024-10-28
bug: /donate giving HTTP/403 in link checker


mcdurdin:
e.g.

```
...
Getting links from: http://localhost:8053/keyboards?q=l:id:mdx-ethi
├─BROKEN─ http://localhost:8053/donate (HTTP_403)
├─BROKEN─ http://localhost:8053/donate (HTTP_403)
├─BROKEN─ http://localhost:8053/donate (HTTP_403)
...
```

https://github.com/keymanapp/keyman.com/actions/runs/11492617843/job/31986970702

Probably relating to URL rewrite update on donate.keyman.com last week?

PR #492, opened 2024-10-31
fix: exclude /donate from broken link checker

mcdurdin:
/donate redirects to another site that gives a 403 forbidden response to bots, which breaks the link checker. So just exclude it.

Fixes: #488

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,6 @@
 name: CI
 
-on: 
+on:
   [pull_request]
 jobs:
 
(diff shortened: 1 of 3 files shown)


darcywong00 (approved):
lgtm assuming the CI link check passes

mcdurdin:
Hmm, turns out that there are 4 references to broken-link-checker. I can understand 2 of them, and no obvious way to consolidate those, but not sure why we have the ones in composer.json as well, would recommend removing them?

darcywong00:
> but not sure why we have the ones in composer.json as well, would recommend removing them?

Ah, I added the one in composer.json in #389 for feature parity with help.keyman.com

I suppose we can remove in a separate PR

mcdurdin:
> Ah, I added the one in composer.json in #389 for feature parity with help.keyman.com

Hmm, so we should probably remove from all then. Let's keep composer.json minimal, remove all scripts, seeing as we do them in ci.yml now?

I see we never call `composer run test` either, which means we should probably add `vendor/bin/phpunit --testdox` to ci.yml and/or Dockerfile?

How do we run lint and phpunit tests locally?

darcywong00:
> I see we never call `composer run test` either, which means we should probably add `vendor/bin/phpunit --testdox` to ci.yml and/or Dockerfile?
> 
> How do we run lint and phpunit tests locally?

Looks like a TODO 😞 
https://github.com/keymanapp/keyman.com/blob/03025d8d5f9843e1ddaca205de059aaa794cf115/build.sh#L31

mcdurdin:
Okay, added #493 to track composer.json/test changes

Issue #384, opened 2023-09-18
feat: Localize keyboard search page


darcywong00:
From the Sept 2023 team planning meetings M4 review of websites

While we would like to localize all the websites, it's too much content (keyman and help.keyman), so the immediate priority is the keyboard search page.

* Get the keyboard search strings into Crowdin (may need to be separate from the Keyman project due to syncing
* Have a way for user to select locale

darcywong00:
Possibly use [docusaurus](https://docusaurus.io/docs/i18n/crowdin) along with crowdin...

darcywong00:
nogo with docusaurus since it needs a static Markdown-generated site.

Now exploring [CakePHP](https://crowdin.com/blog/2021/12/08/cakephp-localization) and .po files

darcywong00:
Using gettext without setting server locale
https://stackoverflow.com/questions/2319060/using-php-gettext-extension-vs-php-arrays-in-multilingual-websites/2319192#2319192

darcywong00:
For adding formatted strings with gettext
https://stackoverflow.com/a/16922727/1836776

Maybe add a wrapper
`_s("welcome_string", $user_name)`   where `welcome_string == "Hello %1$s"`  in the .po

so for the .po, we'd have:
```
# Welcome string for site visitors
#  parameter 1: string, name of site visitor
msgid "welcome_string"
msgstr "Hi there %1$s!"
```

And `_s` would be something like
```php
function _s($s, ...) {
  return sprintf(_($s), ...);
}
```

darcywong00:
From B18S2 Sprint plan meeting, we'll focus on stabilizing other issues for 18.0 beta and resume this work in 19.0

ermshiperete:
See also https://github.com/keymanapp/keyman/issues/9575

PR #500, opened 2024-11-13
feat: Allow keyboard search to be localized 📡

darcywong00:
For starting #384 in localizing the keyboard search page.
(Results come back from  api.keyman.com and currently not localized)

1. The Dockerfile image has updates to install an `n_US.UTF-8` locale. Rather than installing more locales, we'll stick with one and use textdomain() to specify the localized files.

2. Strings contained in .po files and get compiled into .mo files during the `./build.sh start` step. All the .po files need to the in `/_includes/locale/en/LC_MESSAGE/`

3. Add `_s()` wrapper so we can do formatted strings with `_()` - the PHP alias to `gettext()`

4. For now, setup the strings on crowdin at
https://crowdin.com/project/keymancom

- [x] Will move to LTOPs later.

### Usage

For now, the strings can be accessed by adding a parameter e.g.
http://keyman.com.localhost/keyboards/?locale=fr-FR
image

http://keyman.com.localhost/keyboards?locale=es-ES
image


_includes/locale/locale.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+  const locales = [
+    'en'    => 'English', // Default
+    'es-ES' => 'Española',
+    'fr-FR' => 'Français'
+  ];
```
darcywong00, line 7: Currently not used. Maybe when we settle on UX for specifying UI locale on the site...

keyboards/index.php
```diff
@@ -46,14 +56,14 @@
...
+  <h2 class="red underline"><a href='/keyboards'><?= _("Keyboard Search") ?></a></h2>
 
   <div id='search-box'>
     <form method='get' action='/keyboards' name='f'>
-      <label for="search-q">Keyboard search:</label><input id="search-q" type="text" placeholder="Enter language or keyboard" name="q"
+      <label for="search-q"><?= _s("Keyboard search%s", ":") ?></label><input id="search-q" type="text" placeholder="<?= _("Enter language or keyboard")?>" name="q"
       <?php if($embed == 'none') echo 'autofocus'; ?>>
-      <input id="search-f" type="image" src="<?= cdn('img/search-button.png"') ?>" value="Search" onclick="return do_search()">
-      <label id="search-new"><a href='/keyboards<?=$session_query_q?>'>New search</a></label>
+      <input id="search-f" type="image" src="<?= cdn('img/search-button.png"') ?>" value="<?= _("Search") ?>" onclick="return do_search()">
```
darcywong00, line 65: The screenshots show the search button still in English - 
It's because we're using a static image
https://github.com/keymanapp/keyman.com/blob/master/cdn/dev/img/search-button.png

darcywong00: I could convert to a label, if someone can provide some CSS-fu styling

```html
<label id="search-f" class="keyboard_search_button" alt="search button" onclick="return do_search()"> <?= _("Search") ?> </label>
```

darcywong00:
Pausing on this after :duck: with @mcdurdin about the PHP plumbing of localization files/configuration

darcywong00:
Abandoning this PR in favor of #521 onward

PR #515, opened 2024-11-29
fix: use property_exists instead of array_key_exists (continued)

darcywong00:
Follows #513 and replaces the rest of the deprecated references of `array_key_exists` with `property_exists`.

Involves swapping parameters...

### DIFF
--- a/_includes/includes/ui/keyboard-details.php
+++ b/_includes/includes/ui/keyboard-details.php
@@ -94,7 +94,7 @@
 
     protected static function map_license($s) {
       $license_map = ['mit' => 'MIT'];
-      return array_key_exists($s, $license_map) ? $license_map[$s] : $s;
+      return property_exists($license_map, $s) ? $license_map[$s] : $s;
     }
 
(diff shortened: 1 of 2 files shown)


_includes/includes/ui/keyboard-details.php
```diff
@@ -94,7 +94,7 @@ public static function get_tier_from_request($tier = 'stable') {
 
     protected static function map_license($s) {
       $license_map = ['mit' => 'MIT'];
-      return array_key_exists($s, $license_map) ? $license_map[$s] : $s;
+      return property_exists($license_map, $s) ? $license_map[$s] : $s;
```
mcdurdin, line 97: This is incorrect -- `$license_map` is an array not an object

mcdurdin (changes requested):
I am not sure about any of these usages, sorry. I know the license one is incorrect.

darcywong00:
I see. for PHP 7.4 we can still use array_key_exists on arrays. The deprecation notice is on checking objects.

PR #517, opened 2024-12-04
feat: Configure `en_US` locale in Dockerfile 📡

darcywong00:
Some prep work towards #384 of localizing the keyboard search

This brings in the Dockerfile updates from #500 to set up the en_US locale and install the PHP gettext.

The locale/README.md is just capturing an initial setup and will be refined in future PR's.

The .po localization files will also be added on separate PR's and reside in /_includes/locale/en/LC_MESSAGES/

### DIFF
--- a/resources/init-container.sh
+++ b/resources/init-container.sh
@@ -20,7 +20,7 @@ find . -type f -name '*.mo' -delete
 # Compile .po to .mo localization files
 for filename in `find . -type f -name "*.po"`; do
   # Remove the .po extension
-  base_name=`echo ${filename} | sed 's/\.[^.]*$//'`
+  base_name="${filename%.po}"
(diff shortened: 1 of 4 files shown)


resources/init-container.sh
```diff
@@ -10,3 +10,20 @@ else
...
+echo "---- Generating .mo localization files ----"
+cd _includes/locale/en/LC_MESSAGES/
+
+# cleanup previous .mo files
+find . -type f -name '*.mo' -delete
+
+# Compile .po to .mo localization files
+for filename in `find . -type f -name "*.po"`; do
+  # Remove the .po extension
+  base_name=`echo ${filename} | sed 's/\.[^.]*$//'`
```
mcdurdin, line 23:
```suggestion
  base_name="${filename%.po}"
```

mcdurdin (approved):

PR #520, opened 2024-12-09
feat: Add Locale.php class to set current locale 📡

darcywong00:
Continues #384

This starts a Locale.php class which will expand to handle pointing to the localized.po files.
For now, it allows the current locale to be set with a query parameter like `lang=fr-FR` (previously was `locale`)

And confirmed with test code on keyboards/index.php
```php
      <li>Current locale is <?= Locale::currentLocale() ?></li>      
```

### DIFF
--- a/keyboards/session.php
+++ b/keyboards/session.php
@@ -31,8 +31,8 @@
   $embed_android = $embed == 'android';
   $embed_ios = $embed == 'ios';
 
-  if(isset($_REQUEST['locale'])) {
-    \Keyman\Site\com\keyman\Locale::overrideCurrentLocale($_REQUEST['locale']);
+  if(isset($_REQUEST['lang'])) {
+    \Keyman\Site\com\keyman\Locale::overrideCurrentLocale($_REQUEST['lang']);
   }
 
   if($embed != 'none') {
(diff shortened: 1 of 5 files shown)


keyboards/session.php
```diff
@@ -31,6 +31,10 @@
   $embed_android = $embed == 'android';
   $embed_ios = $embed == 'ios';
 
+  if(isset($_REQUEST['locale'])) {
+    \Keyman\Site\com\keyman\Locale::overrideCurrentLocale($_REQUEST['locale']);
+  }
```
mcdurdin, line 36: OK, but we do need to validate that it's an acceptable locale

mcdurdin (approved):

mcdurdin:
I haven't really considered how we set locale via url (keyman.com/en/foo vs keyman.com/foo?lang=en vs en.keyman.com/foo). (I like `lang` instead of `locale` but it's not critical). All of these options have pros and cons... bleagh :)

keyboards/session.php
darcywong00: TODO on follow-on PR once I add some .po files and workout how it fallsback to en strings.

darcywong00:
> I haven't really considered how we set locale via url (keyman.com/en/foo vs keyman.com/foo?lang=en vs en.keyman.com/foo). (I like `lang` instead of `locale` but it's not critical). All of these options have pros and cons... bleagh :)

I changed the query parameter from `locale` to `lang`

PR #526, opened 2025-01-04
fix: avoid error when jsFilename is missing or mismatches package id

mcdurdin:
Ensures that the keymanweb keyboard reference is correct for the package, as it will not always match the package identifier, or may not be available as a standalone keyboard. This then prevents the error in #525 from presenting.

This fix works around a data issue documented in keymanapp/keyman#12853.

I also cleaned up some of the confusion between package id and keyboard id -- but there are deeper ambiguities still present w.r.t package/keyboard id here which we may just need to live with.

Fixes: #525
Relates-to: keymanapp/keyman#12853

### DIFF
--- a/_includes/includes/ui/keyboard-details.php
+++ b/_includes/includes/ui/keyboard-details.php
@@ -55,7 +55,7 @@
 
     /**
      * render_keyboard_details - display keyboard download boxes and details
-     * @param $id - keyboard ID
+     * @param $id - keyboard package ID
      * @param string $tier - ['stable', 'alpha', or 'beta']
      * @param bool $landingPage - when true, details won't display keyboard search box or title
      * @param string $bcp47 - BCP 47 tag to pass as a hint to download links for apps to make connection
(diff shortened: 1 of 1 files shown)


_includes/includes/ui/keyboard-details.php
```diff
@@ -418,12 +431,18 @@ protected static function WriteKeymanWebBox() {
...
+        return;
+      }
+
       // only inject on desktop platforms
       $webDevice = self::GetWebDeviceFromPageDevice();
       if(!$webDevice) {
         return;
       }
 
+      $kmwid = self::GetWebKeyboardId();
```
darcywong00, line 445:
```suggestion
      $keymanWebId = self::GetWebKeyboardId();
```

Don't we want to get away from using `kmw` in our codebase? :smile: 
For l. 474 too

_includes/includes/ui/keyboard-details.php
```diff
@@ -55,7 +55,7 @@ class KeyboardDetails
 
     /**
      * render_keyboard_details - display keyboard download boxes and details
-     * @param $id - keyboard ID
+     * @param $id - keyboard package ID
```
darcywong00, line 58: A chore:
Would `/\$id/\$packageId/` help clear up ambiguity?

darcywong00 (approved):
Nits on cleaning up variable names

_includes/includes/ui/keyboard-details.php
mcdurdin: We have to check each usage so we couldn't just search/replace...

_includes/includes/ui/keyboard-details.php
```diff
@@ -452,7 +471,7 @@ function() {
                 document.getElementById('osk-host').appendChild(newOSK.element);
               }
             );
-            keyman.addKeyboards('<?= self::$id ?>');
+            keyman.addKeyboards('<?= $kmwid ?>');
```
mcdurdin, line 474:
```suggestion
            keyman.addKeyboards('<?= $keymanWebId ?>');
```

PR #536, opened 2025-01-30
feat(developer): support search for keyboards for TIKE project clone

mcdurdin:
Relates-to: keymanapp/keyman#13073

### DIFF
--- a/_includes/includes/ui/keyboard-details.php
+++ b/_includes/includes/ui/keyboard-details.php
@@ -97,6 +97,20 @@
       return array_key_exists($s, $license_map) ? $license_map[$s] : $s;
     }
 
+    protected static function WriteDeveloperCloneBox() {
+      $filename = self::$id . ".kpj";
+      $installLink = 'keyman:keyboard/install/' . rawurlencode(self::$id);
+      if(!empty(self::$bcp47)) $installLink .= "?bcp47=" . rawurlencode(self::$bcp47);
+      $h_filename = htmlspecialchars($filename);
+
+      return <<<END
+<div class="download download-developer">
+  <div class="download-description">Fill in New Project Details box below and click OK to clone $h_filename.</div>
+</div>
+END;
+  // <script>location.href = '$installLink';</script>
+    }
+
     protected static function download_box($platform) {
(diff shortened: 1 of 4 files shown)


_includes/includes/ui/keyboard-details.php
```diff
@@ -361,27 +375,32 @@ protected static function WriteTitle() {
...
+        foreach(self::platformTitles as $platform => $title) {
+          if($platform != 'desktopWeb' && $platform != 'mobileWeb')
+            $text .= self::download_box($platform);
+        }
 
-      $text .= self::download_box('unknown');
+        $text .= self::download_box('unknown');
 
-      if ($embed_win && isset(self::$keyboard->minKeymanVersion) && version_compare(self::$keyboard->minKeymanVersion, $embed_version) > 0) {
+        if ($embed_win && isset(self::$keyboard->minKeymanVersion) && version_compare(self::$keyboard->minKeymanVersion, $embed_version) > 0) {
```
ermshiperete, line 397: `$embed_win` does probably not mean Windows, otherwise I'd wonder why we do that check only for Windows.

ermshiperete (approved):

cdn/dev/keyboard-search/search.css
```diff
@@ -465,6 +465,10 @@ html[data-platform='unknown'] .download.download-unknown {
   display: block;
 }
 
+.download.download-developer {
```
darcywong00, line 468: Could you explain why this doesn't use the pattern `html[data-platform='developer']` like above?

_includes/includes/ui/keyboard-details.php
mcdurdin: `$embed_win` does mean Windows. That's a good question, and I suspect it's because we didn't have Keyman version information being passed in to the site from Linux and mac? Perhaps we should open an issue to explore that in a future version?

cdn/dev/keyboard-search/search.css
mcdurdin: The `html[data-platform]` attribute is the operating system, so it would be `windows` and not `developer`.

PR #546, opened 2025-02-16
fix: redirect to /keyboards if no id provided to /keyboards/install

mcdurdin:
Fixes a 500 server error reported in Google Search Console. Also fixes a TIER_DEVELOPMENT issue with the page where a missing `use KeymanVersion` caused another unrelated 500 error.


darcywong00 (approved):
lgtm

PR #548, opened 2025-02-18
chore: improve sentry reporting

mcdurdin:
* Updates to latest version of Sentry 9.1.0
* Moves Sentry integration into sentry.js
* Ensures that we can capture failed resource loads
* Adds crossorigin for keymanweb.js so we can see failures

- [x] TODO: bootloader v0.18 in build.sh

Depends-on: keymanapp/shared-sites#54
Relates-to: #534


darcywong00 (approved):
lgtm

PR #558, opened 2025-03-12
chore: use simpler bootstrap and common sentry.js

mcdurdin:
Common sentry.js also includes better reporting of missing resources, by attempting a re-fetch of the resource in order to get http status, etc.

Relates-to: keymanapp/shared-sites#55
Relates-to: keymanapp/shared-sites#56
Relates-to: #534

TODO:
- [ ] After shared-sites PR56 merges, update BOOTSTRAP_VERSION


darcywong00 (approved):
lgtm

Also update `BOOTSTRAP_VERSION ` before merging...

_includes/2020/templates/Head.php
```diff
@@ -4,8 +4,14 @@
   namespace Keyman\Site\com\keyman\templates;
 
 use Keyman\Site\com\keyman\Util;
+use Keyman\Site\com\keyman\KeymanComSentry;
 use Keyman\Site\Common\KeymanHosts;
 
+// *Don't* use autoloader here because of potential side-effects in older pages
```
darcywong00, line 10: Does this only apply to the keyman.com site?

mcdurdin: If it isn't there in the other sites, then no.

Issue #561, opened 2025-03-22
bug(web): unable to get Chinese keyboard to display in KeymanWeb Language Demo


Meng-Heng:
Community site: https://community.software.sil.org/t/keymanweb17/9940
> Is it a bug that the keyboard does not change when I select Chinese on the above site?

---
### Location:

Web Page: https://keyman.com/developer/keymanweb/sample2
![Image](https://github.com/user-attachments/assets/760400d2-1611-474c-a162-ef7bc29f5d00)

---
### Issue Analysis:

The Chinese language is using "zh" _(script?)_ as the value but I can't find a Chinese keyboard which uses this script. Could this be the issue since:
* Finnish is using fi as the value and the keyboard has the same layout:
1. KeymanWeb Link: https://keymanweb.com/#fi,Keyboard_basic_kbdfi
![Image](https://github.com/user-attachments/assets/0d312d0c-8cf5-4715-b5eb-7b8a962d003d)

2. KeymanWeb Language Demo:
![Image](https://github.com/user-attachments/assets/c8a06f98-1d09-435c-85f3-596ff3362b56)
---
I can have a look at this further and find a solution if I am applicable to be assigned for this issue?

darcywong00:
On the same community site thread, 

@katelem24 also adds
> I can’t find Obolo too.

My impression is that keymanweb.com is the primary place for users to try out keyboards, and sample2 was never intended to have feature parity.

ermshiperete:
After selecting a language, the sample2 page sends a query like `https://api.keyman.com/cloud/4.0/keyboards?jsonp=keyman.register&languageidtype=bcp47&version=17.0&keyboardid=@zh&timerid=266` with `zh` being the language code. keyman.com will return a JSON object with a list of all the keyboards that support that language. The page then shows the first keyboard from the list. For `zh` there are no keyboards returned, although a search for `l:id:zh` (https://keyman.com/keyboards/languages/zh) lists cs_pinyin (Simplified Chinese) to have the exact bcp47 tag `zh`, and two other related keyboards.

@mcdurdin Is the API not returning cs_pinyin because it's only supported on the desktop platforms?

And what should we do in the sample page (sample2)? The easiest would be to just remove "Chinese" from the list...

ermshiperete:
> I can’t find Obolo too.

Since this is just a sample it doesn't list all languages that Keyman supports, and doesn't provide access to all keyboards either.

mcdurdin:
[...]

PR #565, opened 2025-04-02
fix: remove languages without keyboard and reword placeholder in demo2

ermshiperete:
This change modifies demo2 and removes languages from the list that don't provide a web keyboard. Also reword the placeholder text to make it clearer that it's only a selection of languages.

Fixes: #561


mcdurdin (approved):
LGTM. An alternative would have been to load an additional keyboard by language id (I think we probably do have keyboards for all those languages)

PR #570, opened 2025-04-10
fix: Add `en_app_rgb_wo_60.png` asset to local CDN

darcywong00:
Fixes: #569 
Fixes KEYMAN-COM-1T1

Various users get the error
> Failed to load image https://developer.android.com/images/brand/en_app_rgb_wo_60.png: error 'TypeError: Failed to fetch'

This adds the external asset to local CDN.
Follows AppStore pattern of referencing image


android/index.php
```diff
@@ -12,7 +12,8 @@
     'showMenu' => true,
     'banner' => [
       'title' => 'Type in Your Language On Your Android Device!',
-      'button' => '<a href="'.$playstore.'"><img id="app-store" src="https://developer.android.com/images/brand/en_app_rgb_wo_60.png" alt="Android app on Google Play"></a>',
+      'button' => '<a href="'.$playstore.'"><img id="app-store" src="cdn(en_app_rgb_wo_60.png)"
```
mcdurdin, line 15: This doesn't look right to me? But I see that it is, wow, that's pretty magical and horrible:

https://github.com/keymanapp/keyman.com/blob/a2c22da27126d69a23f8fa7b66b0f94bd75140b0/_includes/includes/template.php#L127-L130

Definite candidate for rewrite in the future!

mcdurdin (approved):

android/index.php
darcywong00: I found that piece of magic from 
https://github.com/keymanapp/keyman.com/blob/9ca42f7f62ffd31ed2e9038f4d0ab992289e2e25/iphone-and-ipad/index.php#L15

mcdurdin: Yeah, I saw that too, and then had to dig because I couldn't quite believe it... not sure who wrote it now, and doesn't really matter, but it's certainly an unnecessarily magical way to solve the problem!

PR #604, opened 2025-09-09
feat: Use associative array for strings on keyboard search page 🗺️

darcywong00:
Replaces the PR chain #516 in addressing #384 of localizing the keyboard search

From the A19S11 sprint design meeting, we pivoted from using .po locale files to localizing strings in associative arrays.

This PR reuses earlier Locale.php work to go in that direction on the main keyboard search page.
For now, locale is passed with the query parameter `lang=` (e.g. `lang=es` or `lang=fr`).

This also adds a globe icon to the top right for selecting UI language (passing the lang parameter)

image

**Limitation:** Currently session is set in keyboards/session.php so the rest of the site is ignoring the lang query parameter

TODO's on follow-on PRs
- [x] keyboard search results (Details and Install pages) :link: #622
- [x] check that crowdin.com can sync with these .php files (previously was syncing .po files)

## User Testing
**Setup** - Run the keyman.com and website-local-proxy sites locally with docker

* **TEST_ENGLISH** - Verifies default strings
1.  Navigate to http://keyman.com.localhost/keyboards
2. Verify English strings match the live site on https://keyman.com/keyboards

* **TEST_SPANISH** - Verifies search page appear in Spanish
1. Navigate to http://keyman.com.localhost/keyboards
2. Hover on the globe key in the top right (next to the rest donate section) and select *Spanish - Español*
3. Verify the main keyboard search page displays in Spanish

image

* **TEST_FRENCH** - Verifies search page appear in French
1.  Navigate to http://keyman.com.localhost/keyboards?lang=fr-FR
2. Hover on the globe key in the top right (next to the rest donate section) and select *French - Français*
3. Verify keyboard search page is in French

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,10 +37,11 @@ jobs:
     - name: Check broken links
       shell: bash
       continue-on-error: false
+      # Exclude checking locales for each link 'lang=*'
       run: |
         set +e
         set +o pipefail
-        npx broken-link-checker http://localhost:8053/_test --recursive --ordered ---host-requests 50 -e --filter-level 3 --exclude '*/donate' | tee blc.log
+        npx broken-link-checker http://localhost:8053/_test --recursive --ordered ---host-requests 50 -e --filter-level 3 --exclude '*/donate' --exclude '*lang=*' | tee blc.log
         echo "BLC_RESULT=${PIPESTATUS[0]}" >> "$GITHUB_ENV"
 
     - name: Report on broken links
--- a/build.sh
+++ b/build.sh
@@ -42,8 +42,9 @@ function test_docker_container() {
   docker exec $KEYMAN_CONTAINER_DESC sh -c "find . -name '*.php' | grep -v '/vendor/' | xargs -n 1 -d '\\n' php -l"
 
   # NOTE: link checker runs on host rather than in docker image
+  # Also exclude testing locales for each link: 'lang=*'
   builder_echo blue "---- Testing links"
-  npx broken-link-checker http://localhost:8053/_test --recursive --ordered ---host-requests 50 -e --filter-level 3 --exclude '*/donate' | tee blc.log
+  npx broken-link-checker http://localhost:8053/_test --recursive --ordered ---host-requests 50 -e --filter-level 3 --exclude '*/donate' --exclude '*lang=*' | tee blc.log
   local BLC_RESULT=${PIPESTATUS[0]}
   echo ----------------------------------------------------------------------
   echo Link check summary


_includes/locale/Locale.php
```diff
@@ -0,0 +1,112 @@
...
+      'en',
+      'es-ES',
+      'fr-FR'
+    );
+
+    // xx-YY locale as specified in crowdin %locale%
+    private static $currentLocale = Locale::DEFAULT_LOCALE;
+
+    // If set to true, warn about untranslated strings
+    private static $debugMode = false; // Keep false in production
```
mcdurdin, line 24: Can't this be deduced from the environment?

_includes/locale/Locale.php
```diff
@@ -0,0 +1,112 @@
...
+  namespace Keyman\Site\com\keyman;
+
+  class Locale {
+    public const DEFAULT_LOCALE = 'en';
+
+    public const CROWDIN_LOCALES = array(
+      'en',
+      'es-ES',
+      'fr-FR'
+    );
```
mcdurdin, line 18: I think this should be in a separate file, or perhaps determined on a per-domain basis, because we have multiple l10n domains (i.e. keyboards is one localization domain). See my comments re filenames and folder names further down

_includes/locale/Locale.php
```diff
@@ -0,0 +1,112 @@
...
+
+    /**
+     * Validate $locale is an acceptable locale.
+     * Using xx-YY as specified in crowdin %locale%
+     * @param $locale - the locale to validate
+     * @return true if valid locale
+     */
+    public static function validateLocale($locale) {
+      return in_array($locale, Locale::CROWDIN_LOCALES);
+    }
```
mcdurdin, line 56: This may not be viable because we may not have localizations for a given l10n domain.

_includes/locale/Locale.php
```diff
@@ -0,0 +1,112 @@
...
+
+    // If set to true, warn about untranslated strings
+    private static $debugMode = false; // Keep false in production
+
+    // Associative array to hold translated strings in "current locale" and English
+    private static $langArray = [];
+    private static $langArrayEn = [];
+
+    /**
+     * Return the current locale. Fallback to 'en'
```
mcdurdin, line 31: Fallback should really be through a path, e.g. given `es-ES`, we should fallback first to `es`, then to `en`.

_includes/locale/Locale.php
```diff
@@ -0,0 +1,112 @@
...
+
+  namespace Keyman\Site\com\keyman;
+
+  class Locale {
+    public const DEFAULT_LOCALE = 'en';
+
+    public const CROWDIN_LOCALES = array(
+      'en',
+      'es-ES',
+      'fr-FR'
```
mcdurdin, line 17: These should be minimal BCP-47 tags - i.e. we would use `es-419` for Latin America and Caribbean, `es` for Spain. Similar for French, should be just `fr`, not `fr-FR` (we would use `fr-CA` for example).

_includes/locale/Locale.php
```diff
@@ -0,0 +1,112 @@
...
+
+    // xx-YY locale as specified in crowdin %locale%
+    private static $currentLocale = Locale::DEFAULT_LOCALE;
+
+    // If set to true, warn about untranslated strings
+    private static $debugMode = false; // Keep false in production
+
+    // Associative array to hold translated strings in "current locale" and English
+    private static $langArray = [];
+    private static $langArrayEn = [];
```
mcdurdin, line 28: Can we find a better name than `langArray`? `strings`?

_includes/locale/Locale.php
```diff
@@ -0,0 +1,112 @@
...
+     */
+    public static function currentLocale() {
+      return Locale::$currentLocale;
+    }
+
+    /**
+     * Validate and override the current locale
+     * @param $locale - the new current locale (xx-YY as specified in crowdin %locale%)
+     */
+    public static function overrideCurrentLocale($locale) {
```
mcdurdin, line 42: `overrideCurrentLocale` should probably be `setLocale`. This is a global setting for the site. Here, we set the primary locale, with an array of fallbacks, ending always in 'en'.

```php
function setLocale($locale) {
  // TODO: support other fallbacks such as es-419->es
  // for now, just support locale and 'en'
  Locale::$currentLocales = [$locale, 'en'];
}
```

So we don't do validation here, because we'll always have fallback anyway.

_includes/locale/Locale.php
```diff
@@ -0,0 +1,112 @@
...
+      return in_array($locale, Locale::CROWDIN_LOCALES);
+    }
+
+    /**
+     * Reads localized strings from the specified $domain-locale.php file
+     * @param $domain - base filename of the .php file containing localized strings
+     * (name not including -xx-YY locale)
+     * path is relative to _includes/locale/
+     */
+    public static function localize($domain) {
```
mcdurdin, line 64: Can we call this `loadStrings()` or something like that?

_includes/locale/Locale.php
```diff
@@ -0,0 +1,112 @@
...
+      $currentLocaleFilename = sprintf("%s/%s-%s",
+        __DIR__ . '/en/LC_MESSAGES/',
```
mcdurdin, line 66: I don't think we should use a path `en/LC_MESSAGES` and then put localizations all in a single folder. I would go:

```
_includes/locale/strings/
  keyboards/
    en.php
    es.php
  home
    en.php
    es.php
    km.php
```

We will need to be able to load multiple domains in the same page - e.g. `template` (i.e. header and footer) and `keyboards`. So we need to refactor this class to support loading strings for multiple domains. Given what I noted about fallback earlier, I think we should be considering something like this (this is pseudocode):

```php
// strings is an array of domains, each domain is an array of locales, each locale
// is an object? with loaded flag and array of strings.
private static $strings = [];

function loadDomain($domain) {
  $strings[$domain] = [];
  $files = glob(/locale/strings/$domain);
  if(failed) return false;
  for($file in $files) {
    $file = stripExtension($file);
    $strings[$domain][$file] = { strings: [], loaded: false };
  }
  return true;
}
```

Then the string loading should be on demand on first use, again pseudocode:

```php
function getString($domain, $id) {
  if(!array_has_key($domain, $strings)) {
    if(!self::loadDomain($domain)) die('no such domain');
  }
  for($locale in self::currentLocales()) {
    if(!array_has_key($strings[$domain], $locale)) {
      continue;
    }
    if(!$strings[$domain][$locale]->loaded) {
      // will set ->loaded = true
      self::loadStrings($domain, $locale);
    }
    if(array_has_key($strings[$domain][$locale]->strings, $id)) {
      return $strings[$domain][$locale]->strings[$id];
    }
  }
  // string not found in any localization oops
  if(dev_mode / debug / !production) { // need the right test here
    die(`string $id is missing in all l10ns`);
  }
  return $id;
}
```

_includes/locale/Locale.php
```diff
@@ -0,0 +1,112 @@
...
+      } else {
+        die("English strings in $enLocaleFilename not found");
+      }
+    }
+
+    /**
+     * Wrapper to lookup string. Fallback to English
+     * @param $id - the key
+     */
+    public static function _m($id) {
```
mcdurdin, line 91: If this is a member of `Locale` class, there's no need to use the `_`. Perhaps just `m`? But: see `_s` discussion, plus we'll need to pass `$domain` in here too

_includes/locale/Locale.php
```diff
@@ -0,0 +1,112 @@
...
+      return self::$langArray[$id];
+    }
+
+    /**
+     * Wrapper to format string with gettext '_(' alias and variable args
+     * Placeholders should escape like %1\$s
+     * @param $s - the format string
+     * @param $args - optional remaining args to the format string
+     */ 
+    public static function _s($s, ...$args) {
```
mcdurdin, line 109: Ditto on no `_` needed. I would have the one public function though (and make `_m` into a private `getString()`):

```php
public static function m($domain, $id, ...args) {
  $str = self::getString($domain, $id);
  if(count($args) == 0) return $str;
  return vsprintf($str, $args);
}
```

_includes/locale/Locale.php
```diff
@@ -0,0 +1,112 @@
...
+
+    // Associative array to hold translated strings in "current locale" and English
+    private static $langArray = [];
+    private static $langArrayEn = [];
+
+    /**
+     * Return the current locale. Fallback to 'en'
+     * @return $currentLocale
+     */
+    public static function currentLocale() {
```
mcdurdin, line 34: This should return an array of available locales in priority order, e.g. for `es-ES`, we'd go `['es-ES', 'es', 'en']`. This way, we can support fallback more neatly in the future, even if we initially only go with one fallback to `en`.

mcdurdin (changes requested):
This is a nice clean start. But I am asking for a bunch of changes to make it more extensible long-term. Thanks!

_includes/locale/Locale.php
darcywong00: Can you give an example? 
We can set the list of available languages on
https://crowdin.com/project/keymancom/settings#languages

(would need the sil_ltops login to add custom languages)

_includes/locale/Locale.php
darcywong00: with the glob refactor below, this can go away

_includes/locale/Locale.php
darcywong00: removed with the glob refactor.

_includes/locale/Locale.php
darcywong00: So should setLocale() split on dashes for determining fallback.
e.g. with `es-ES` we end up with
```php
self::$currentLocales = ['es-ES', 'es', 'en'];
```

mcdurdin: There are algorithms for fallback already. See https://cldr.unicode.org/development/development-process/design-proposals/language-distance-data

That might be too much for us. @srl295 may have suggestions on simpler approaches.

darcywong00: Moved to `calculateFallbackLocales()` where we can always reimplement in a TODO

keyboards/index.php
```diff
@@ -7,10 +7,13 @@
   use Keyman\Site\com\keyman\templates\Menu;
   use Keyman\Site\com\keyman\templates\Body;
   use Keyman\Site\com\keyman\templates\Foot;
+  use Keyman\Site\com\keyman\Locale;
+
+  Locale::localize('keyboards');
```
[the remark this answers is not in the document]
darcywong00: Will it get messy if every page we localize has to add?
```php
function_m($id) {...}
```

or should we just move it to Localize.php()?

darcywong00:
I think this is ready for another round of review comments.

This one is still unresolved.
https://github.com/keymanapp/keyman.com/pull/604#discussion_r2332090428
> > It would be good to look at best practices around embedding locale in query strings vs path or domain etc. We need to consider our technology stack in making this decision.

For now, I'm just using `?lang=es` or `?lang=fr` for testing.

keyboards/session.php
```diff
@@ -32,6 +32,13 @@
   $embed_ios = $embed == 'ios';
   $embed_developer = $embed == 'developer';
 
+  if(isset($_REQUEST['lang'])) {
+    \Keyman\Site\com\keyman\Locale::overrideCurrentLocale($_REQUEST['lang']);
+  } else if (isset($_SESSION['lang'])) {
+    \Keyman\Site\com\keyman\Locale::overrideCurrentLocale($_SESSION['lang']);
+  }
+  $_SESSION['lang'] = \Keyman\Site\com\keyman\Locale::currentLocale();
+
```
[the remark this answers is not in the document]
darcywong00: From SO: "Does the locale belong on the path or as a request parameter on the URL?
https://stackoverflow.com/a/16471122

> If you need it as request parameter or part of the url depends of what you want to achieve. If you want to serve static content, you should have it be part of the path. If you want to act dynamically on the chosen locale, you should use it as request parameter, since you don't want to have your scripts replicated several times over different paths just to add different locales.

darcywong00:
Hmm. something to chase down on why the link checker is now 7x longer
Reference GHA
https://github.com/keymanapp/keyman.com/actions/runs/19664278223/job/56317324603?pr=624


| PR | Links Found | Links Excluded | Elapsed Time |
|-----|--------------------|-----------------------|---------------------|
| #750 (baseline) | 318,228 | 315,763 | 21 minutes, 24 seconds |
| This PR | 2,377,385 | 2,366,016 | 1 hour, 50 minutes |

_includes/2020/templates/Menu.php
```diff
@@ -29,6 +29,52 @@ public static function render(array $fields): void {
...
+    }
+
+    /**
+     * Render the globe dropdown for changing the UI language.
+     * As UI languages get added, we'll need to update this
+     */
+    private static function render_globe_dropdown(): void {
+?>
+          <p>
+            <div id="ui-language" class="menu-item">
```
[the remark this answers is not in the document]
darcywong00: Updated to pass a div number so we can assign "id" = `ui-language` or `ui-language1`

Meng-Heng:
### Prerequisites
1. Fetch & Pull `feat/localize/keyboards-map`
2. Run `keyman.com` & `website-local-proxy` on Chrome

# Test Results

* **TEST_ENGLISH (PASSED):**
1. Open `http://keyman.com.localhost:8053/keyboards/`
2. Verified: The page content is in English & match the content on `keyman.com/keyboards`.

* **TEST_SPANISH (PASSED):**
1. Return to Home page, in English
2. Open `http://keyman.com.localhost:8053/keyboards/`, the page content remains in English.
3. Hover the globe key, click on Español
4. Verified: The page content translates to Spanish.

* **TEST_FRENCH (PASSED):**
1. Return to Home page, in Spanish
2. Open `http://keyman.com.localhost:8053/keyboards/`, the page content is still in Spanish
3. Hover the globe key, click on Français
4. Verified: The page content translates to French
5. Return to Home page, in French
6. Open `http://keyman.com.localhost:8053/keyboards/`, the page content remains in French
7. Hover on the globe key, click on English
8. Open `http://keyman.com.localhost:8053/keyboards/?lang=fr-Fr`
9. Verified: the page opens up in French.

PR #613, opened 2025-10-07
feat: rewrite Location header to https for live sites

mcdurdin:
Testing this on keyman-staging.com before going live.

Fixes: keymanapp/keyman#14328
Test-bot: skip


darcywong00:
btw, I don't think we've reverted keymanapp/keyman#7227 yet, so the pre-release apps are still using the production sites, right?

PR #624, opened 2025-11-25
maint: emit broken-link-checker logs before summarizing

mcdurdin:
Test-bot: skip


mcdurdin:
If there are broken links, they show up in a separate block so you don't get overwhelmed by the passing links:

image

PR #626, opened 2025-12-03
Limit globe button visibility to `/keyboards/` pages 🗺️

darcywong00:
Addresses this point of #384

> We should hide the icon on pages that don't have localization (to avoid disappointment for users)

Will limit visibility to `/keyboards/` pages

Test-bot: skip


mcdurdin (approved):
LGTM 😉

PR #658, opened 2026-02-12
feat: localize JS keyboard search 🗺️

darcywong00:
For #384 

Most of the PHP keyboard search pages can be localized.
This follows on by allowing the JS keyboard search to also be localized.

Off-the-shelf, [i18n-next](https://www.i18next.com/) looked really good, but I opted for something that didn't need a framework:
https://medium.com/@mihura.ian/translations-in-vanilla-javascript-c942c2095170

## Some of the changes

The JSON files make it easy to
* add strings
* add formatted parameters to strings
For example:
```javascript
t('pageNumberOfTotalPages', {pageNumber: res.context.pageNumber, totalPages: res.context.totalPages})
```

I had to rename `search.js` --> `search.mjs` and and add `i18n.mjs` as modules in order to import the JSON files.

## Sample screenshots
french sil

french more

Test-bot: skip


cdn/dev/keyboard-search/search.js
```diff
@@ -258,25 +260,25 @@ function process_response(q, obsolete, res) {
...
       if(isKeyboardObsolete(kbd) && !deprecatedElement) {
         // TODO: make title change depending on whether deprecated keyboards are shown or hidden
         deprecatedElement = $(
-          '<div class="keyboards-deprecated"><h4 class="red underline">Obsolete keyboards</h4></div>');
+          '<div class="keyboards-deprecated"><h4 class="red underline">' + t('obsoleteKeyboards') + '</h4></div>');
         resultsElement.append(deprecatedElement);
       }
 
-      $keyboardClass = kbd.isDedicatedLandingPage ? 'keyboard keyboardLandingPage' : 'keyboard';
+      var keyboardClass = kbd.isDedicatedLandingPage ? 'keyboard keyboardLandingPage' : 'keyboard';
```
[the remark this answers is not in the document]
mcdurdin: It would have worked before being a es module -- it would have effectively been a property of `window`, `window.$keyboardClass`. But you cannot use variables without declaration in es module.

```suggestion
      const keyboardClass = kbd.isDedicatedLandingPage ? 'keyboard keyboardLandingPage' : 'keyboard';
```

Nothing to do with jQuery. Using `$` suggests it might have been originally in some PHP code.

cdn/dev/js/i18n/es.json
```diff
@@ -0,0 +1,54 @@
+{
+  "resultOne": {
+    "text": "resultado",
+    "crowdinContext": "1 keyboard result found"
```
mcdurdin: I don't like the crowdinContext being repeated for each language -- this is a lot of busy noise.

Couldn't we just have an object map?

```json
{ 
  "resultOne": "resultado", 
  ... 
}
```

cdn/dev/js/i18n/i18n.mjs
```diff
@@ -0,0 +1,111 @@
+/**
+ * Keyman is copyright (c) SIL Global. MIT License
+ * 
+ * Vanilla JS for localizing keyboard search strings without a framework
+ * Reference: https://medium.com/@mihura.ian/translations-in-vanilla-javascript-c942c2095170
+ */
+import translationEN from './en.json' with { type: 'json' };
+import translationES from './es.json' with { type: 'json' };
+import translationFR from './fr.json' with { type: 'json' };
```
mcdurdin: This is not scalable to many translations. We need to use dynamic imports, and see also my comment on en.json - the translations need to be scoped.

cdn/dev/js/i18n/en.json
```diff
@@ -0,0 +1,54 @@
+{
```
mcdurdin: This won't scale, because it's just one json file, so strings are going to get all muddled together. We are going to need to split per .js file I think?

cdn/dev/js/i18n/i18n.mjs
```diff
@@ -0,0 +1,111 @@
...
+    if (!I18n.translations[language]) {
+      // Langage is missing, so fallback to "en"
+      console.warn(`i18n for language: '${language}' missing, fallback to 'en'`);
+      language = "en";
+    }
+
+    if (!I18n.translations[language].hasOwnProperty(key)) {
+      // key is missing for current language
+      console.warn(`key '${key}' missing in '${language}' translations`);
+      return '';
```
mcdurdin: Fallback to `en` string would be better because we know we have those.

mcdurdin (changes requested):

cdn/dev/js/i18n/en.json
darcywong00: > We are going to need to split per .js file I think?

So the strings relating to `search.mjs` would go in a folder like this?
`/cdn/dev/js/i18n/search/`en.json, es.json, etc?

_legacy/keyboards/index.php
```diff
@@ -39,4 +39,4 @@
   <p>Loading...</p>
 </div>
 
-<script src="<?= cdn('legacy-keyboard-search/search.js'); ?>"></script>
\ No newline at end of file
+<script src="<?= cdn('legacy-keyboard-search/search.mjs'); ?>" type="module"></script>
```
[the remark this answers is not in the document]
darcywong00: ahh.... reverted

darcywong00:
Replaced by #715

PR #659, opened 2026-02-12
fix: make section announcement function static

mcdurdin:
Fixes: KEYMAN-COM-321
Fixes: KEYMAN-COM-322
Test-bot: skip


mcdurdin:
> Why does it have to be static, or why do we want it to be static?

We are just calling it as a static method. We don't need to instantiate the class because it has no data. The class is just wrapping it for simpler namespace-based usage.

PR #662, opened 2026-02-14
feat: add Peirce Baehr as former team member

mcdurdin:
Test-bot: skip


mcdurdin:
> Do we have permission on the profile pic? (includes children)
> Or should we crop it closer?

Yes, that was the pic he suggested, it's from Peirce's FB profile. I've asked him if the cropping is good, will wait on his reply.

PR #668, opened 2026-02-24
chore: rework i18n for js to avoid async and duplication

mcdurdin:
Moves the responsibility for determining locale fallback server-side, so that we can avoid async pollution and have just a single responsible module for locale fallback.

Strings are injected into a <script> data element as JSON, reducing the number of network round trips.

@darcywong00 happy to discuss this further as required

Test-bot: skip


darcywong00 (approved):
nice - this does make things cleaner!

PR #703, opened 2026-03-30
chore: A19S25 merge master into i18n-url-scheme 🌋

darcywong00:
End of sprint merge.

CI link check test will fail cause the epic branch still over-hauling directories

Test-bot: skip
Build-bot: build


mcdurdin (approved):
RSLGTM

nit: `auto` should only be used for actual automatically-generated and managed PRs. This should be `chore` unless it ran automatically

PR #704, opened 2026-03-31
fix: Start to incorporate lang in URL paths 🌋

darcywong00:
Follows #696

This overhauls .htaccess files to handle a langtag in the URL path (rewrite to corresponding  _content/ path).
Flow of .htaccess:
1. Redirections before rewriting
2. Remove file extensions from links
3. Known path redirections
4. i18n redirect --> `/file` to `<lang>/file`
5. Rewrite rules

Additional small changes:
* Moved ‎`_content/archive/.htaccess` rule to the root .htaccess
* cleanup downloads path
* cleanup some _test page links

CI link checker ~will~ may still fail

Test-bot: skip


.htaccess
```diff
@@ -32,6 +32,14 @@ RewriteRule ^crowdin.yml$ = [F,L]
 # apple-app-site-association
 RewriteRule "^.well-known/apple-app-site-association$" "/.well-known/apple-app-site-association.json" [L]
 
+# Redirect /archive/downloads.php"
+RewriteRule "^(archive)/(downloads)(\.php)?$" "en/$2/$1/" [NC,R=301,QSA]
+
+# Rewrite /<lang>/... ignoring known 2-3 letter folders
+RewriteCond "$1" !-f
+RewriteCond "$1" !-d
+RewriteRule "^(?!cdn|go|mac|web)([^/]+)/(.+)$" "/_content/$2?lang=$1"
```
darcywong00: It feels like we'd have to split this into many cases like the following?
$2.php
$2/index.php
$2/index.md

darcywong00: Handled by reorganizing the .htaccess flow:
1. Redirections - before rewriting (includes removing file extensions)
2. Rewrites (to _content as needed)
3. PHP and Markdown rewriting

_content/android/app/index.php
```diff
@@ -0,0 +1,4 @@
+<?php
+    // Redirect to Android homepage
+    header('Location: https://help.keyman.com/products/android/');
```
darcywong00: Converted from `/_content/android/app/.htaccess` - also done for iphone-and-ipad/app

Root .htaccess rules weren't matching

.htaccess
```diff
@@ -24,14 +24,49 @@ RewriteBase /
...
 
 # Deny crowdin.yml
-RewriteRule ^crowdin.yml$ = [F,L]
+RewriteRule "^crowdin.yml$" = [F,L]
 
 # Custom error messages (need to pass the original request)
-ErrorDocument 404 /_includes/errors/404
+ErrorDocument 404 /_includes/errors/404.php
+
+RewriteRule "^_includes/errors/404(.php)?$" "/_includes/errors/404.php" [END]
```
mcdurdin: I think we should avoid serving pages from _includes directly, so perhaps this should be /404.php or /errors/404.php?

.htaccess
```diff
@@ -24,14 +24,49 @@ RewriteBase /
...
+# ------------------------------------------------------------------
+
+# Remove index.md or index.php and redirect (and stop processing)
+#RewriteCond "$1" -d
+RewriteRule "^((.+)/)?index(.md|.php)?$" "$1" [R,L]
+
+# Remove .php extension and redirect
+#RewriteCond "$1.php" -f
+#RewriteCond "$1" !-d
+RewriteRule "^(.+).php$" "$1" [R,L]
```
mcdurdin: We need the same pattern for Markdown I think

```suggestion
RewriteRule "^(.+)\.php$" "$1" [R,L]
RewriteRule "^(.+)\.md$" "$1" [R,L]
```

See for example:

http://keyman.com.localhost/en/about/team/bios/darcywong00.md should redirect to http://keyman.com.localhost/en/about/team/bios/darcywong00

.htaccess
```diff
@@ -24,14 +24,49 @@ RewriteBase /
...
+# ##################################################################
+# Redirections - before rewriting
+# ##################################################################
+
+# ------------------------------------------------------------------
+# PHP and Markdown redirections
+# ------------------------------------------------------------------
+
+# Remove index.md or index.php and redirect (and stop processing)
+#RewriteCond "$1" -d
```
mcdurdin:
```suggestion
```

.htaccess
```diff
@@ -24,14 +24,49 @@ RewriteBase /
...
+RewriteRule "^((.+)/)?index(.md|.php)?$" "$1" [R,L]
+
+# Remove .php extension and redirect
+#RewriteCond "$1.php" -f
+#RewriteCond "$1" !-d
```
mcdurdin:
```suggestion
```

.htaccess
```diff
@@ -24,14 +24,49 @@ RewriteBase /
...
+# Redirections - before rewriting
+# ##################################################################
+
+# ------------------------------------------------------------------
+# PHP and Markdown redirections
+# ------------------------------------------------------------------
+
+# Remove index.md or index.php and redirect (and stop processing)
+#RewriteCond "$1" -d
+RewriteRule "^((.+)/)?index(.md|.php)?$" "$1" [R,L]
```
mcdurdin:
```suggestion
RewriteRule "^((.+)/)?index(\.md|\.php)?$" "$1" [R,L]
```

.htaccess
```diff
@@ -24,14 +24,49 @@ RewriteBase /
...
+RewriteCond "{DOCUMENT_ROOT}/$1.php" !-f
+RewriteCond "{DOCUMENT_ROOT}/$1.md" !-f
+RewriteRule "^(.+[^/])$" "$1/" [R,END]
+
+# ------------------------------------------------------------------
+# Known path redirections
+# ------------------------------------------------------------------
+
+# Redirect /archive/downloads.php"
+RewriteRule "^archive/downloads(.php)?$" "/downloads/archive/" [NC,R=301,QSA]
```
mcdurdin:
```suggestion
RewriteRule "^archive/downloads(.php)?$" "/downloads/archive/" [NC,R=301,END,QSA]
```

? I am not sure if `END` is required for redirects?

.htaccess
```diff
@@ -180,52 +163,147 @@ RedirectMatch "^/(french|german|italian|spanish|swedish)(/?)$"       "/keyboards
...
+RewriteRule "^(en)/downloads/releases/(.+)$" "/_content/downloads/releases/_version_downloads.php?version=$2&lang=$1" [END]
 
 #
-# PHP rewriting
+# Assets don't use lang param
+#
+RewriteCond "%{DOCUMENT_ROOT}/cdn/$1$2" -f
+RewriteRule "^cdn/(.+\.)(gif|js|png|svg)$" "/cdn/$1$2" [END]
+
+RewriteRule "^(en)/(.+\.)(gif|js|png|svg)$" "/_content/$2$3" [END]
```
mcdurdin: This is a bit fragile - what if we have other file types? jpeg etc? Can we not just use:

```suggestion
RewriteCond "%{DOCUMENT_ROOT}/$2" -f
RewriteRule "^(en)/(.+)$" "/_content/$2" [END]
```

.htaccess
```diff
@@ -180,52 +163,147 @@ RedirectMatch "^/(french|german|italian|spanish|swedish)(/?)$"       "/keyboards
...
+
+# releases-download
+RewriteRule "^(en)/downloads/releases/(.+)$" "/_content/downloads/releases/_version_downloads.php?version=$2&lang=$1" [END]
 
 #
-# PHP rewriting
+# Assets don't use lang param
+#
+RewriteCond "%{DOCUMENT_ROOT}/cdn/$1$2" -f
+RewriteRule "^cdn/(.+\.)(gif|js|png|svg)$" "/cdn/$1$2" [END]
```
mcdurdin: This should be with the _control/go/_test rules

_content/android/app/index.php
mcdurdin: I'm a little uncomfortable with this change -- it might be better to add the rule at the top level .htaccess?

_content/downloads/index.php
```diff
@@ -29,5 +29,5 @@
 
 
 <?php
-  require_once('./_downloads.php');
+  require_once('_downloads.php');
```
mcdurdin: Why this change? The `./` is important for preventing search along the path

_content/keyboards/h/tamil/index.php
```diff
@@ -102,13 +102,13 @@
   <h3 id="cta-other-downloads">Download a Tamil keyboard on these devices:</h3>
 
   <div class="download-cta-small iPhone" id="cta-iPhone">
-    <a href=../../../iphone/">
+    <a href="../../../../iphone/">
```
mcdurdin: This change seems wrong?

_content/keyboards/h/tamil/index.php
```diff
@@ -102,13 +102,13 @@
...
       <img src="<?php echo cdn('img/cta-icons/icon-iphone.png'); ?>" />
       <p>iPhone</p>
     </a>
   </div>
   <div class="download-cta-small iPad" id="cta-iPad">
-    <a href="../../../ipad/">
+    <a href="../../../../ipad/">
```
mcdurdin: Ditto?

_content/keyboards/h/urdu/index.php
```diff
@@ -103,7 +103,7 @@
     </a>
   </div>
   <!--<div class="download-cta-small Bookmarklet" id="cta-Bookmarklet">
-    <a href="../../../bookmarklet/">
+    <a href="../../../../bookmarklet/">
```
mcdurdin:
```suggestion
    <a href="../../../bookmarklet/">
```

_content/windows/index.php
```diff
@@ -168,6 +168,6 @@
     <span class="red">Q.</span> Are Keyman Desktop 7.1 and other previous releases still available for download?
 </p>
 <p>
-    <span class="red">A.</span> Yes, you can find the links to download Keyman Desktop 7.1 and other previous releases <a href="../archive/downloads.php">here</a>.
+    <span class="red">A.</span> Yes, you can find the links to download Keyman Desktop 7.1 and other previous releases <a href="/archive/downloads">here</a>.
```
mcdurdin:
```suggestion
    <span class="red">A.</span> Yes, you can find the links to download Keyman Desktop 7.1 and other previous releases <a href="../archive/downloads">here</a>.
```

Why make the path absolute again?

_includes/2020/templates/Foot.php
```diff
@@ -12,11 +12,6 @@ static function render(array $fields = []) {
       $fields = (object)$fields;
       $fields->stable_version = KeymanVersion::stable_version;
       $fields->beta_version = KeymanVersion::beta_version;
-
-      // Fallback to 'en'
-      // TODO-I18N-URL-SCHEME: integrate with Locale.php
-      $fields->lang = isset($_REQUEST['lang']) ? $_REQUEST['lang'] : 'en';
```
mcdurdin, line 18: Why remove this? We'll need something like this throughout

_test/index.md
```diff
@@ -23,15 +23,15 @@ showmenu: false
 
 ## Links to Other Test Pages
 
-[/go links](./go)
+[/go links](/_test/go)
```
mcdurdin: Why this change?

mcdurdin (changes requested):

_content/windows/index.php
darcywong00: This depends on the root level redirect
https://github.com/keymanapp/keyman.com/blob/11c3a701aeba7bf14828c9d543b28a668f4335f5/.htaccess#L64-L65

When I make it relative path, I get a 404

Image

Since this is our own link, maybe I should just fix it to `../downloads/archive`

_test/index.md
darcywong00: The CI link checker test was failing
https://github.com/keymanapp/keyman.com/actions/runs/24058950224/job/70171010949#step:7:18

I don't quite understand it, but the CI link checker would treat these links as:
```bash
Getting links from: http://localhost:8053/_test
├─BROKEN─ http://localhost:8053/go (HTTP_403)
├─BROKEN─ http://localhost:8053/language-landing (HTTP_404)
├─BROKEN─ http://localhost:8053/legacy (HTTP_404)
├─BROKEN─ http://localhost:8053/well-known (HTTP_404)
Error: Process completed with exit code 1.
```

_content/keyboards/h/tamil/index.php
darcywong00: good catch. these two should match android/bookmarklet.

_content/downloads/index.php
darcywong00: ic, will revert

_content/android/app/index.php
darcywong00: ok, will convert this too top level .htaccess rule(s)

_includes/2020/templates/Foot.php
darcywong00: ah, I got mixed up thinking the new .htaccess rules would handle the language tag.

So we want to keep $fields->lang for  Foot, Menu?

_includes/2020/templates/Foot.php
```diff
@@ -65,8 +60,8 @@ static function render(array $fields = []) {
         </div>
         <div class="footer-third sil-logo">
             <br>
-            <a href="/<?=$fields->lang?>/about/"><img id="sil-logo" src="<?php echo ImageRandomizer::randomizer(); ?>" width="50%" alt='SIL' /></a>
-            <p>Created by <a href="/<?=$fields->lang?>/about/">SIL Global</a></p>
+            <a href="/about/"><img id="sil-logo" src="<?php echo ImageRandomizer::randomizer(); ?>" width="50%" alt='SIL' /></a>
+            <p>Created by <a href="/about/">SIL Global</a></p>
```
darcywong00: btw, why doesn't about SIL Global go to https://global.sil.org/about/ ?

_content/windows/index.php
mcdurdin: Definitely we should fix the link on our site, no point in additional complexity and redirects.

But this highlights an upcoming issue with redirects which I think we kinda decided punt until later: what do we do with things like keyman.com/khmer --> /keyboards/khmer_angkor? Do we also support keyman.com/en/khmer --> /en/keyboards/khmer_angkor? Do we do this for all redirects or a subset or none?

I am unsure.

_includes/2020/templates/Foot.php
mcdurdin: Yes, we'll want to keep those fields but they'll need to be extracted from the URL. I think we'll be able to parse the URL as it should be the first path component, rather than using URL rewriting to append a lang parameter?

_test/index.md
mcdurdin: This means that the rule which appends a / in a redirect for index.md is missing or out of order; we'll need that.

- /_test should redirect to /_test/ because it is a folder.

alongside the rules that strip .php and .md suffix and deal with `index` filename:

- /_test/index.php -> /_test/index
- /_test/index.md -> /_test/index
- /_test/index -> /_test/

_includes/2020/templates/Foot.php
mcdurdin: Yeah, probably should!

_test/index.md
darcywong00: k, still troubleshooting this one.
Will get back to this after Songkran break!

_content/windows/index.php
```diff
@@ -168,6 +168,6 @@
     <span class="red">Q.</span> Are Keyman Desktop 7.1 and other previous releases still available for download?
 </p>
 <p>
-    <span class="red">A.</span> Yes, you can find the links to download Keyman Desktop 7.1 and other previous releases <a href="../archive/downloads.php">here</a>.
+    <span class="red">A.</span> Yes, you can find the links to download Keyman Desktop 7.1 and other previous releases <a href="../downloads/archive">here</a>.
```
darcywong00, line 171: This link is cleaned up

_includes/2020/templates/Menu.php
```diff
@@ -296,12 +296,12 @@ private static function render_top_menu(object $fields): void {
...
-                            <li><a href="/<?=$fields->lang?>/<?=$fields->lang?>/free/">Free on all Platforms</a></li>
-                            <li><a href="/<?=$fields->lang?>/<?=$fields->lang?>/ldml/">LDML Support</a></li>
-                            <li><a href="<<?=$fields->lang?>/?= KeymanHosts::Instance()->help_keyman_com ?>">Help and Documentation</a></li>
-                            <li><a href="/<?=$fields->lang?>/contact/">Contact Us</a></li>
+                            <li><a href="/<?=$fields->lang?>/about/get-involved">Get Involved</a></li>
+                            <li><a href="/<?=$fields->lang?>/training">Training Events</a></li>
+                            <li><a href="/<?=$fields->lang?>/free/">Free on all Platforms</a></li>
+                            <li><a href="/<?=$fields->lang?>/ldml/">LDML Support</a></li>
+                            <li><a href="<?= KeymanHosts::Instance()->help_keyman_com ?>">Help and Documentation</a></li>
+                            <li><a href="/contact/">Contact Us</a></li>
```
mcdurdin:
```suggestion
                            <li><a href="<?=$fields->lang?>/contact/">Contact Us</a></li>
```

_test/index.md
darcywong00: aha, `{DOCUMENT_ROOT}` needed to be `%{DOCUMENT_ROOT}`

fixed in https://github.com/keymanapp/keyman.com/pull/704/commits/88769e9e847cdd5f679571b6432181e746e9854e along with
* cleaning up links in _test
* cleaning up unncessary `\/` to `/` in .htaccess

_includes/2020/templates/Menu.php
darcywong00: We had leading slashes in the previous links
```suggestion
                            <li><a href="/<?=$fields->lang?>/contact/">Contact Us</a></li>
```

mcdurdin: 🤦🏼 oops good catch!

_includes/2020/templates/Foot.php
darcywong00: Handled in #711

_content/windows/index.php
darcywong00: atm
keyman.com/khmer --> /en/keyboards/khmer_angkor

keyman.com/en/khmer --> 404

mcdurdin: I think this is _probably_ okay given we never link to these directly on our site?

EDIT: my bad, the homepage links directly to /khmer ... so we lose language choice if we link directly?

mcdurdin (approved):

PR #717, opened 2026-04-30
fix: handle redirects for paths and files with terminating slashes 🌋

mcdurdin:
Remove the `/` from a path if it refers to a file, and add a `/` if the path refers to a folder (implicit index.md/index.php).

Clean up keyboard search which had some assumptions about not including the `/` but that doesn't make sense with this pattern.

Test-bot: skip

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -23,16 +23,106 @@ RewriteBase /
 # Add mime type for apple-app-site-association
 AddType application/json apple-app-site-association
 
-# Deny crowdin.yml
-RewriteRule "^crowdin.yml$" = [F,L]
-
 # Custom error messages (need to pass the original request)
 ErrorDocument 404 /_includes/errors/404.php
 
 RewriteRule "^_includes/errors/404(.php)?$" "/_includes/errors/404.php" [END]
 
+# ##################################################################
+# Reject special and legacy top-level files without i18n
+# ##################################################################
+
+# Any existing file in _common/assets - must come first because it is overridden below otherwise
+RewriteCond "%{DOCUMENT_ROOT}/$1/$2" -f
+RewriteRule "^(_common/assets)/(.+)$" - [END]
+
+#
+# The following top-level files and folders are not directly visible through URLs:
+#
+#    _common/      (excl. _common/assets/)
+#    _content/
+#    _includes/
+#    _scripts/
+#    .github/
+#    resources/
+#    tests/
+#    .editorconfig
+#    .gitattributes
+#    .gitignore
+#    .htaccess      (this file, not listed explicitly)
+#    build.sh
+#    composer.json
+#    composer.lock
+#    crowdin.yml
+#    Dockerfile
+#    package-lock.json
+#    package.json
+#    phpunit.xml
+#    README.md
+#    TODO.md
+#
+RewriteRule "^(_common|_content|_includes|_scripts|.github|resources|tests)(/.*|$)" - [F,END]
+RewriteRule "^(.editorconfig|.gitattributes|.gitignore|build.sh|composer.json|composer.lock|crowdin.yml|Dockerfile|package-lock.json|package.json|phpunit.xml|README.md|TODO.md)$" - [F,END]
+
+#
+# The following paths are accessible but controlled by sub-folder .htaccess:
+#    _legacy/
+# Thus this rule currently has no effect but will come into play if the
+# _legacy/.htaccess file is removed.
+#
+RewriteRule "^(_legacy)(/.*|$)" - [F,END]
+
+# ##################################################################
+# Handle special and legacy top-level files without i18n
+# ##################################################################
+
+#
+# The following top-level files and folders do not have i18n redirection,
+# but have .php and .md rewrites. They do not have sub-folders.
+#    _common/assets/    (handled above because _common/ itself is excluded)
+#    _control/
+#    _ie_thunk/
+#    _test/
+#
+# The following top-level files and folders do not have .php or .md rewrites
+#    .well-known/
+#    cdn/
+#    go/
+#    robots.txt
+#    tier.txt
+#
+
 # apple-app-site-association
 RewriteRule "^.well-known/apple-app-site-association$" "/.well-known/apple-app-site-association.json" [L]
+
+# Add terminating slash on top-level folders by redirecting
+RewriteCond "$1" ^(_control|_ie_thunk|_test|.well-known|go|cdn)$
+RewriteRule "^([^/]+)$" "$1/" [R,L]
+
+# .php rewrite
+RewriteCond "$1" ^(_control|_ie_thunk|_test)$
+RewriteCond "%{DOCUMENT_ROOT}/$1/$2.php" -f
+RewriteRule "^([^/]+)/(.+)$" "/$1/$2.php" [END]
+
+# .md rewrite
+RewriteCond "$1" ^(_control|_ie_thunk|_test)$
+RewriteCond "%{DOCUMENT_ROOT}/$1/$2.md" -f
+RewriteRule "^([^/]+)/(.+)$" "/_includes/includes/md/mdhost.php?file=$1/$2.md" [END]
+
+# .md rewrite for folder; no .php rewrite for folder
+RewriteCond "$1" ^(_control|_ie_thunk|_test)$
+RewriteCond "%{DOCUMENT_ROOT}/$1/index.md" -f
+RewriteRule "^([^/]+)/$" "/_includes/includes/md/mdhost.php?file=$1/index.md" [END]
+
+# Any existing file in any of those folders or sub folders
+RewriteCond "$1" ^(_control|_ie_thunk|_test|.well-known|go|cdn)$
+RewriteCond "%{DOCUMENT_ROOT}/$1/$2" -f
+RewriteRule "^([^/]+)/(.+)$" - [END]
+
+# Top-level files
+RewriteCond "$1" ^(robots.txt|tier.txt)$
+RewriteCond "%{DOCUMENT_ROOT}/$1" -f
+RewriteRule "^([^/]+)$" - [END]
 
 # ##################################################################
 # Redirections - before rewriting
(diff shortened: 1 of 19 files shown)


.htaccess
```diff
@@ -49,17 +49,64 @@ RewriteRule "^((.+)/)?index(\.md|\.php)?$" "$1" [R,L]
...
+
+# Redirect folder to folder/ for .md
+RewriteCond "%{DOCUMENT_ROOT}/$1" -d
+RewriteCond "%{DOCUMENT_ROOT}/$1/index.md" -f
+RewriteRule "^(.+[^/])$" "$1/" [R,L]
+
+# Redirect file/ to file (.md)
+RewriteCond "%{DOCUMENT_ROOT}/$1" !-d
+RewriteCond "%{DOCUMENT_ROOT}/$1.md" -f
+RewriteRule "^(.+)/$" "$1" [R,L]
```
darcywong00, line 190: not sure why CI link checker is now flagging
```
Getting links from: http://localhost:8053/_test
├─BROKEN─ http://localhost:8053/_test/keyboards (BLC_UNKNOWN)
```

Test page link:
https://github.com/keymanapp/keyman.com/blob/3471e043d3dd12dd81177800df2887693b4d3847/_test/index.md?plain=1#L28

mcdurdin: Ah yes, I saw this -- it's because the rules only apply to files in _content right now. We might need a special set of rules to handle those? Or perhaps we can simplify by prepending the locale before any other rules, and moving go, _test, and any other addressable content to _content?

_content/keyboards/h/amharic/index.php
```diff
@@ -96,13 +96,13 @@
   <h3 id="cta-other-downloads">Download an Amharic keyboard on these devices:</h3>
 
   <div class="download-cta-small iPhone" id="cta-iPhone">
-    <a href="../../../iphone/">
+    <a href="../../../iphone-and-ipad/">
```
mcdurdin, line 99: We keep /ipad and /iphone as top-level redirects, but we don't need those for e.g. /en/ipad as the content is all at /_content/iphone-and-ipad. This eliminates a rule in the .htaccess

_content/keyboards/index.php
```diff
@@ -59,14 +61,14 @@ function _m($id, ...$args) {    return Locale::m(LOCALE_KEYBOARDS, $id, ...$args
...
-  <h2 class="red underline"><a href='/<?= $head_options['language']?>/keyboards'><?= _m('page_title') ?></a></h2>
+  <h2 class="red underline"><a href='<?=$keyboardsPage?>'><?= _m('page_title') ?></a></h2>
 
   <div id='search-box'>
-    <form method='get' action='/<?= $head_options['language']?>/keyboards' name='f'>
+    <form method='get' action='<?=$keyboardsPage?>' name='f'>
       <label for="search-q"><?= _m('keyboard_search') ?></label><input id="search-q" type="text" placeholder="<?= _m('enter_language') ?>" name="q"
       <?php if($embed == 'none') echo 'autofocus'; ?>>
       <input id="search-f" type="button" value="<?= _m('search') ?>">
-      <label id="search-new"><a href='/<?= $head_options['language']?>/keyboards<?=$session_query_q?>'><?= _m('new_search')?></a></label>
```
mcdurdin, line 69: This line was broken (nested `<?=`), when I fixed it I decided to simplify all references

.htaccess
```diff
@@ -184,104 +319,95 @@ RewriteRule "^(downloads/releases|keyboards)/(.+)$" "en/$1/$2" [R,L]
...
-RewriteRule "^(_control|go|_test)/(.+)$" "/$1/$2.php" [END]
-
-RewriteCond "%{DOCUMENT_ROOT}/$1/$2.md" -f
-RewriteRule "^(_control|go|_test)/(.+)$" "/_includes/includes/md/mdhost.php?file=$1/$2.md" [END]
-
-RewriteCond "%{DOCUMENT_ROOT}/$1/index.md" -f
-RewriteRule "^(_control|go|_test)(/?)$" "/_includes/includes/md/mdhost.php?file=$1/index.md" [END]
-
-RewriteCond "%{DOCUMENT_ROOT}/cdn/$1$2" -f
-RewriteRule "^cdn/(.+\.)(gif|js|png|svg)$" "/cdn/$1$2" [END]
```
mcdurdin, line 203: This section is now moved up

.htaccess
```diff
@@ -184,104 +319,95 @@ RewriteRule "^(downloads/releases|keyboards)/(.+)$" "en/$1/$2" [R,L]
...
 #
 
 #
 # keyboards Install | Download | Share | bare | .json -->
 #
 
 # /keyboards/install/[id] to /keyboards/install.php
-RewriteRule "^(([a-z]{2,3})(-([A-Za-z]{4}))?(-([a-z]{2}|[0-9]{3}))?)/keyboards/install/([^/]+)$" "/_content/keyboards/install.php?id=$7&lang=$1" [NC,END,QSA]
+RewriteCond "$1" ^([a-z]{2,3})(-([A-Za-z]{4}))?(-([a-z]{2}|[0-9]{3}))?$ [NC]   # BCP 47 match
+RewriteRule "^([^/]+)/keyboards/install/([^/]+)$" "/_content/keyboards/install.php?id=$2&lang=$1" [END,QSA]
```
mcdurdin, line 339: Hmm, was going to put these into a separate PR ... but was working on them and pushed accidentally. Let's see how they go!

.htaccess
```diff
@@ -23,17 +23,105 @@ RewriteBase /
...
+#    crowdin.yml
+#    Dockerfile
+#    package-lock.json
+#    package.json
+#    phpunit.xml
+#    README.md
+#    TODO.md
+#
+RewriteRule "^(_common|_content|_includes|_scripts|.github|resources|tests)(/.*|$)" - [F,END]
+RewriteRule "^(.editorconfig|.gitattributes|.gitignore|build.sh|composer.json|composer.lock|crowdin.yml|Dockerfile|package-lock.json|package.json|phpunit.xml|README.md|TODO.md)$" [F,END]
```
darcywong00: Does this need `- [F,END]` like in the other lines?

.htaccess
```diff
@@ -23,17 +23,105 @@ RewriteBase /
...
+RewriteCond "%{DOCUMENT_ROOT}/$1/$2.md" -f
+RewriteRule "^([^/]+)/(.+)$" "/_includes/includes/md/mdhost.php?file=$1/$2.md" [END]
+
+# .md rewrite for folder; no .php rewrite for folder
+RewriteCond "$1" ^(_control|_test|_ie_thunk)$
+RewriteCond "%{DOCUMENT_ROOT}/$1/index.md" -f
+RewriteRule "^([^/]+)/$" "/_includes/includes/md/mdhost.php?file=$1/index.md" [END]
+
+# Any existing file in any of those folders or sub folders
+RewriteCond "$1" ^(_control|_ie_thunk|_test|.well-known|go|cdn)$
```
darcywong00: To help me with future maintenance, can we have the RewriteConds in the same sort order as the commented folder listing (ll 80-88)?

e.g.
```.htaccess
RewriteCond "$1" ^(_control|_ie_thunk|_test)$
```
and
```.htaccess
RewriteCond "$1" ^(_control|_ie_thunk|_test|.well-known|cdn|go)$
```

.htaccess
```diff
@@ -49,17 +137,64 @@ RewriteRule "^((.+)/)?index(\.md|\.php)?$" "$1" [R,L]
 RewriteRule "^(.+)\.php$" "$1" [R,L]
 RewriteRule "^(.+)\.md$" "$1" [R,L]
 
-# Redirect folder without / to include /
+# ------------------------------------------------------------------
+# Add and remove `/` for folders and files for consistency
```
darcywong00: maybe wordsmithing something like
```suggestion
# For trailing '/' consistency: Add '/' for folders and remove `/` for files
```

darcywong00 (approved):
ahh, nice cleanup! thanks

_content/keyboards/index.php
```diff
@@ -40,6 +40,8 @@ function _m($id, ...$args) {    return Locale::m(LOCALE_KEYBOARDS, $id, ...$args
   if($embed == 'none')
     Menu::render([]); // we'll be doing client-side os detection now
   Body::render();
+
+  $keyboardsPage = '/' . $head_options['language'] . '/keyboards';
```
darcywong00: Devin asks if $keyboardPage missing a trailing slash causes unnecessary 302 redirects.

> New .htaccess rules at lines 62-64 redirect `/en/keyboards` -> `/en/keyboards/` because `_content/keyboards` is a directory containing index.php
> ...
> When a user presses Enter in the search box, this causes an unnecessary redirect from `/en/keyboards?q=...` to `/en/keyboards/?q=...`

.htaccess
mcdurdin: For sure, was aiming for alphabetical but moved some things around and missed those. Will cleanup

cdn/dev/keyboard-search/search.mjs
mcdurdin, line 24: This is not new to this PR, so will handle this separately. But, as devin.ai notes:

cdn/dev/keyboard-search/search.mjs:R22-31

**Module scope prevents search.mjs from reading PHP-set global variables**

The PHP template at `_content/keyboards/index.php:46-48` sets global variables `embed`, `embed_query`, and `embed_lang` via a regular `<script>` tag. However, `search.mjs` is loaded as an ES module (via `import` at `_content/keyboards/index.php:77`), so its `var` declarations are module-scoped and shadow any globals. Specifically, `embed_query` (line 22-24 of search.mjs) always becomes `''` because the hoisted `var` is `undefined` before the assignment. Similarly, `embed_lang` (line 31) is computed from the URL regex rather than the PHP value. The global `embed` variable IS accessible since the module doesn't redeclare it. This is pre-existing behavior unaffected by this PR, but it means the PHP-set `embed_query` and `embed_lang` values are effectively dead code for the module's logic.

mcdurdin: * #720

darcywong00 (approved):
lgtm

PR #712, opened 2026-04-23
refactor: Update globe button to modify `lang` in URL paths 🌋

darcywong00:
Part of #696

Since i18n redirects `/file` to `<lang>/file`, this updates the globe key button to modify the BCP-47 language tag in the URL. The initial implementation of the globe key set the language tag as a query parameter.

Additional changes:
* Replaces `en` in .htaccess with a *basic* regex to allow language-script-region BCP-47 triplets. To keep things simple, intentionally not using the full-blown regex in https://github.com/keymanapp/keyman.com/blob/b448efc6e7751c7ff70dc72b5f28dc40373f84ba/_includes/2020/Validation.php#L28-L37
* Globe key changes:
  *  Update to always be visible (previously implemented to only appear on pages that contained localization)
  * Change phone menu list description from "Keyboard Search UI" to "Display in:"
  * Add `(Khmer)` to language name

### Screenshots

phone layout
Screenshot_2026-04-28_11-08-15

desktop layout
Screenshot_2026-04-28_11-08-37

Test-bot: skip

PR #719, opened 2026-05-05
test: automate determining non-en locales for broken-link-checker test to ignore 🌋

darcywong00:
Addresses https://github.com/keymanapp/keyman.com/pull/712#discussion_r3154723856

> Excluding all locales except en will not scale well. I wonder if there is a better way -- can we build a scripted filter using the api published at [broken-link-checker](https://www.npmjs.com/package/broken-link-checker)?

I'm applying to the base branch master so we can compare the CI link checker to existing runs
This [passing run took 13m 23s](https://github.com/keymanapp/keyman.com/actions/runs/25424339340) after skipping /downloads/releases

The one for [#979 took 26m 49s](https://github.com/keymanapp/keyman.com/actions/runs/25269524878/job/74089491806)

This updates ci.yml and build.sh to determine the list of non-en locales to ignore (along with /downloads/releases) for the broken-link-checker CLI to run.

Test-bot: skip

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,8 +42,8 @@ jobs:
         set +o pipefail
         readarray -t ignoresArray <<< $(find ./_includes/locale/strings/keyboards/ -maxdepth 1 -name '*.php' ! -name "en.php" \
           -execdir basename  {} .php ';')
-        local baseURL="http://localhost:8053"
-        local ignoreStr=("  --exclude ${baseURL}*/downloads/releases/*")
+        baseURL="http://localhost:8053"
+        ignoreStr=("  --exclude ${baseURL}*/downloads/releases/*")
         for locale in "${ignoresArray[@]}"; do
           ignoreStr+=" --exclude ${baseURL}/$locale/*"
         done
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,11 +21,6 @@ jobs:
       env:
         fail-fast: true
 
-    - name: Install node dependencies
-      shell: bash
-      run: |
-        npm install
-
     #
     # Finally, run the tests
     #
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -37,11 +42,11 @@ jobs:
     - name: Check broken links
       shell: bash
       continue-on-error: false
-      # Exclude checking locales for each link 'lang=*'
+      working-directory: ./_test
       run: |
         set +e
         set +o pipefail
-        npx broken-link-checker http://localhost:8053/_test --recursive --ordered ---host-requests 50 -e --filter-level 3 --exclude '*/donate' --exclude '*lang=*' | tee blc.log
+        node blc_test.mjs | tee ../blc.log
```
darcywong00: An alternative (less drastic) way may be to use BASH to generate the list of excludes

like
```suggestion
  # determine non-en locales to ignore
  readarray -t ignoresArray <<< $(find ./_includes/locale/strings/keyboards/ -maxdepth 1 -name '*.php' ! -name "en.php" \
    -execdir basename  {} .php ';')
  local baseURL="http://localhost:8053"
  local ignoreStr=()
  for locale in "${ignoresArray[@]}"; do
    ignoreStr+=" --exclude ${baseURL}/$locale/*"
  done
  # echo "ignoreStr: ${ignoreStr[@]}"
  npx broken-link-checker ${baseURL}/_test --recursive --ordered ---host-requests 50 -e --filter-level 3 ${ignoreStr} | tee blc.log
```

mcdurdin: Yep, I reckon this would be easier for now but I think the .mjs would be more maintainable long-term. Not sure if we have bandwidth to complete the long-term approach.

(Another area we could exclude is the /downloads/releases for every single release)

darcywong00: split the long-term suggestion to #721

mcdurdin (approved):
LGTM

Issue #720, opened 2026-05-06
bug: search.mjs cannot read embed flags from global scope


mcdurdin:
> This is not new to this PR, so will handle this separately. But, as devin.ai notes:
> 
> cdn/dev/keyboard-search/search.mjs:R22-31
> 
> **Module scope prevents search.mjs from reading PHP-set global variables**
> 
> The PHP template at `_content/keyboards/index.php:46-48` sets global variables `embed`, `embed_query`, and `embed_lang` via a regular `<script>` tag. However, `search.mjs` is loaded as an ES module (via `import` at `_content/keyboards/index.php:77`), so its `var` declarations are module-scoped and shadow any globals. Specifically, `embed_query` (line 22-24 of search.mjs) always becomes `''` because the hoisted `var` is `undefined` before the assignment. Similarly, `embed_lang` (line 31) is computed from the URL regex rather than the PHP value. The global `embed` variable IS accessible since the module doesn't redeclare it. This is pre-existing behavior unaffected by this PR, but it means the PHP-set `embed_query` and `embed_lang` values are effectively dead code for the module's logic. 

 _Originally posted by @mcdurdin in [#717](https://github.com/keymanapp/keyman.com/pull/717/changes#r3193702170)_

darcywong00:
Also from https://github.com/keymanapp/keyman.com/pull/715#discussion_r3193877459
```js
var langMatch = location.pathname.match(/(\/(.+))?\/keyboards\/(.*)$/);
var embed_lang = (langMatch) ? langMatch[2] : 'en';
```
> We should extract this into a common module -- as a follow-up PR

mcdurdin:
Fixed by #725

PR #725, opened 2026-05-07
feat: use page URL path to find current locale 🌋

mcdurdin:
* Use first component of URL path to determine current locale
* Add PHP `Locale::pageLocale()` to retrieve the current locale
* Add JS `I18n.pageLocale()` to retrieve the current locale
* Redirect to /en/... if an unsupported locale is found (in head.php)
* Remove `lang` parameter from embedded keyboard search
* Remove `lang` parameter from rewrite rules
* Rename `Locale::definePageLocale()` to `Locale::definePageScope()` to clarify purpose of the function
* Remove lang property from session; we may want to track lang in the future so we remember for top-level redirects
* And fix the embed access in search.mjs (#720)

TODO: Once this change lands, we'll need to rebuild the lang parameter into the `/go/<platform>/download-keyboards` rewrite rules and redirect to the correct locale that way.

Fixes: #720
Test-bot: skip

### DIFF
--- a/_includes/locale/Locale.php
+++ b/_includes/locale/Locale.php
@@ -30,19 +30,57 @@
     // Each locale is an object? with loaded flag and array of strings
     private static $strings = [];
 
+    private static $invalidLocale = false;
+
     /**
      * Return the current locales. Fallback to 'en'
      * @return $currentLocales
      */
     public static function currentLocales() {
+      if(count(self::$currentLocales) == 0) {
+        // For pages with no locale set, or if called too early ...?
+        // var_dump(self::$currentLocales);
+        Locale::setLocaleFromURL();
+      }
       return self::$currentLocales;
+    }
+
+    /**
+     * Return the user-selected page locale, which is always embedded at the
+     * front of the URL path
+     */
+    public static function pageLocale() {
+      return self::currentLocales()[0];
+    }
+
+    public static function invalidLocale() {
+      return self::$invalidLocale;
+    }
+
+    /**
+     *
+     */
+    private static function setLocaleFromURL() {
+      // First component of the URL is always the
+      if(preg_match('/^\\/(([a-z]{2,3})(-([A-Za-z]{4}))?(-([a-z]{2}|[0-9]{3}))?)\\//', $_SERVER['REQUEST_URI'], $matches)) {
+        if(!isset(DISPLAY_NAMES[$matches[1]])) {
+          // Note: this is an unsupported locale, so we'll end up redirecting in head.php to /en/...
+          $pageLocale = Locale::DEFAULT_LOCALE;
+          self::$invalidLocale = true;
+        } else {
+          $pageLocale = $matches[1];
+        }
+      } else {
+        $pageLocale = Locale::DEFAULT_LOCALE;
+      }
+      self::setLocale($pageLocale);
     }
 
     /**
      * Set the current locales, with an array of fallbacks, ending in 'en'.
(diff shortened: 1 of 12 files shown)


_includes/locale/Locale.php
```diff
@@ -30,19 +30,57 @@ class Locale {
...
+    /**
+     *
+     */
+    private static function setLocaleFromURL() {
+      // First component of the URL is always the
+      if(preg_match('/^\\/(([a-z]{2,3})(-([A-Za-z]{4}))?(-([a-z]{2}|[0-9]{3}))?)\\//', $_SERVER['REQUEST_URI'], $matches)) {
+        if(!isset(DISPLAY_NAMES[$matches[1]])) {
+          // Note: this is an unsupported locale, so we'll end up redirecting in head.php to /en/...
+          $pageLocale = Locale::DEFAULT_LOCALE;
+          self::$invalidLocale = true;
```
darcywong00, line 76: "invalidLocale" seems a little fuzzy when:
* specified locale - fallback to 'en'. But we set invalidLocale = true
* invalid locale e.g. 'notavalidtag' - fallback to 'en' and set invalidLocale = true

_includes/locale/Locale.php
```diff
@@ -30,19 +30,57 @@ class Locale {
...
+        } else {
+          $pageLocale = $matches[1];
```
darcywong00, line 78: Should we reset invalidLocale = false?

cdn/dev/keyboard-search/search.mjs
```diff
@@ -19,16 +19,12 @@ if (!String.prototype.includes) {
...
 }
 
-var embed_query_q = embed_query == '' ? '' : '?'+embed_query;
-var embed_query_x = embed_query == '' ? '' : '&'+embed_query;
-
-// TODO: Validate BCP-47 triplet?
-var langMatch = location.pathname.match(/(\/(.+))?\/keyboards\/(.*)$/);
-var embed_lang = (langMatch) ? langMatch[2] : 'en';
+var embed_query_q = window.embed_query == '' ? '' : '?'+window.embed_query;
+var embed_query_x = window.embed_query == '' ? '' : '&'+window.embed_query;
```
darcywong00, line 27: Do we want PHP and JS setting `window.embed_query`?

how is
```suggestion
var embed_query = (typeof window.embed_query != 'undefined') ? window.embed_query : '';

var embed_query_q = embed_query == '' ? '' : '?'+embed_query;
var embed_query_x = embed_query == '' ? '' : '&'+embed_query;
```

_includes/locale/Locale.php
mcdurdin: I think e.g. /notavalidtag/keyboards should 404. But /ka-ZZ/keyboards should redirect to /en/keyboards. This way, /cs-CZ/keyboards when given from the Android app will redirect to /en/keyboards until we have a localization for cz on the site.

(We may want to work on redirecting to a fallback as well, so /en-au/ redirects to /en/, but that should be a follow-up; it can be done in just head.php and Locale.php.)

_includes/locale/Locale.php
mcdurdin: No, it will already be `false` from line 33. `setLocaleFromURL` will only be called once by `currentLocales()` (it's private)

cdn/dev/keyboard-search/search.mjs
mcdurdin: Yeah, it's a failsafe at this point, I don't think it hurts either way.

darcywong00 (approved):
lgtm

PR #733, opened 2026-05-11
fix: setLocaleFromURL handle fallback locals and regex of BCP-47 Script/Region 🌋

darcywong00:
Found while integrating Keyman for Android which reports Spanish locale as `es-ES`

Had to make the following changes to setLocaleFromURL():
* check with fallback locales

* Also update BCP-47 regex to capitalize Script and Region
 
This way, the following link shows in Spanish (which uses es.php files)
http://keyman.com.localhost/es-ES/keyboards/?embed=android&embed_version=19.0

invalid locale like
http://keyman.com.localhost/rv-tS/keyboards/?embed=android&embed_version=19.0

get a page not found.

I couldn't invalidate the locale because that broke _test pages like /_test/go (would treat _test as an invalid locale)

Test-bot: skip


darcywong00:
force rebased onto staging

> yeah, we can rebase to staging and close the web-i18n epic branch, so that we don't have confusion. When ready, we just make a back-port to master from staging PR

_includes/locale/Locale.php
```diff
@@ -69,13 +69,14 @@ public static function invalidLocale() {
      */
     private static function setLocaleFromURL() {
       // First component of the URL is always the locale
-      if(preg_match('/^\\/(([a-z]{2,3})(-([A-Za-z]{4}))?(-([a-z]{2}|[0-9]{3}))?)\\//', $_SERVER['REQUEST_URI'], $matches)) {
-        if(!isset(DISPLAY_NAMES[$matches[1]])) {
+      if(preg_match('/^\/(([a-z]{2,3})(-([A-Za-z]{4}))?(-([a-z]{2}|[0-9]{3}))?)\//i', $_SERVER['REQUEST_URI'], $matches)) {
```
darcywong00: > And we don't want case insensitivity either; 

Then maybe we fix the regex for upper-case regions
```suggestion
      if(preg_match('/^\/(([a-z]{2,3})(-([A-Za-z]{4}))?(-([A-Z]{2}|[0-9]{3}))?)\//', $_SERVER['REQUEST_URI'], $matches)) {
```

mcdurdin: Yep, do we do the same for .htaccess?

mcdurdin: And script should be `(-([A-Z][a-z]{3}))`

darcywong00:
```suggestion
      if(preg_match('/^\/(([a-z]{2,3})(-([A-Z][a-z]{3}))?(-([A-Z]{2}|[0-9]{3}))?)\//', $_SERVER['REQUEST_URI'], $matches)) {
```

darcywong00: .htaccess has the same regex to update. And currently using [NC] for case-insensitive.
Will update those too

PR #730, opened 2026-05-11
fix: Update /go/android links for BCP-47 🌋

darcywong00:
With the revamp of applying BCP-47 tags to keyman.com URLs
I found I needed to modify the /go/android/ links to insert the BCP-47 tag

### language search
http://keyman.com.localhost/go/android/es/19.0/download-keyboards/languages/km
yielded

Screenshot_2026-05-11_09-17-54

### keyboard search
http://keyman.com.localhost/go/android/es/19.0/download-keyboards

Screenshot_2026-05-11_09-18-59

Will need to :cherries: pick to epic/i18n-url-scheme

Test-bot: skip

PR #734, opened 2026-05-12
fix: rewrite /keyboards for embed mode and skip localization

mcdurdin:
Keyman 14.0 - 18.0 have a dependency on /keyboards which means that we cannot rewrite URLs there when we are in 'embed' mode for those apps. The apps check if the URL starts with /keyboards, and handle those paths internally, while opening other URLs in an external browser.

See https://github.com/keymanapp/keyman/issues/15948 for platform-specific dependencies on this behaviour and for v19.0 planned mitigations.

Also remove _ie_thunk and _legacy from .htaccess as these are no longer required following removal of pre-Keyman-13.0 support.

Relates-to: keymanapp/keyman#15948
Test-bot: skip

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -28,6 +28,85 @@ ErrorDocument 404 /_includes/errors/404.php
 
 RewriteRule "^_includes/errors/404(.php)?$" "/_includes/errors/404.php" [END]
 
+# ##############################################################################
+#
+# Keyman 14.0 - 18.0 have a dependency on /keyboards which means that we cannot
+# rewrite URLs there when we are in 'embed' mode for those apps. The apps check
+# if the URL starts with /keyboards, and handle those paths internally, while
+# opening other URLs in an external browser.
+#
+# See https://github.com/keymanapp/keyman/issues/15948 for platform-specific
+# dependencies on this behaviour and for v19.0 planned mitigations.
+#
+# These rules are duplicated for the locale-aware versions for web browser
+# visitors later in this file.
+#
+# ##############################################################################
+
+#
+# keyboards Install | Download | Share | bare | .json -->
+#
+
+# /keyboards/install/[id] to /keyboards/install.php
+RewriteCond "%{HTTP_COOKIE}"    embed_keyboards_no_locale_redirect [OR]
+RewriteCond "%{QUERY_STRING}"   embed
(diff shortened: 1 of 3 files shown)


cdn/dev/keyboard-search/search.mjs
```diff
@@ -294,9 +303,9 @@ function process_response(q, obsolete, res) {
         "</div>");
 
       if(kbd.isDedicatedLandingPage) {
-        $('.title a', k).text(kbd.name).attr('href', `/${I18n.pageLocale()}/keyboards/h/${kbd.id}${embed_query_q}`);
+        $('.title a', k).text(kbd.name).attr('href', `${page_root}/h${kbd.id}${embed_query_q}`);
```
mcdurdin, line 306: Note deletion of extra `/` after `h` which was previously incorrect, because the dedicated landing pages always have a preceding `/`:

https://github.com/keymanapp/keyman.com/blob/b95a4a6b1e1553e7b84ac16a8870c4a517cefb44/cdn/dev/keyboard-search/dedicated-landing-pages.js#L5-L7

.htaccess
```diff
@@ -28,6 +28,85 @@ ErrorDocument 404 /_includes/errors/404.php
...
+# ##############################################################################
+
+#
+# keyboards Install | Download | Share | bare | .json -->
+#
+
+# /keyboards/install/[id] to /keyboards/install.php
+RewriteCond "%{HTTP_COOKIE}"    embed_keyboards_no_locale_redirect [OR]
+RewriteCond "%{QUERY_STRING}"   embed
+RewriteRule "^keyboards/install/([^/]+)$" "/_content/keyboards/install.php?id=$1" [END,QSA]
```
mcdurdin, line 53: The pattern here is similar through this section:
1. Add the two `RewriteCond` lines (`[OR]` flag reduces number of rules!)
2. Strip the BCP47 component from the path
3. Adjust `$2` references to `$1`.

_content/keyboards/session.php
darcywong00, line 48: Do we need to delete the cookie?

darcywong00 (approved):
lgtm.

For 19.0 apps, would we tweak /go/  links in #730?

_content/keyboards/session.php
mcdurdin: No, it's a session cookie, so will expire by itself when the session ends.

PR #736, opened 2026-05-13
fix: Check embed for "new search" and dedicated landing pages

darcywong00:
Follows the `page_root` pattern in #734 for the "New Search" link.
For Keyman 14.0-18.0 apps using the embedded view, we don't include the locale for "Near Search"

Also added a section in .htaccess so dedicated landing pages can check the `embed` session cookie.
Fortunately, all the landing pages are `index.php`.

Test-bot: skip


mcdurdin (approved):
LGTM, I am not sure how the /h/ pages slipped through because I did test those...

mcdurdin:
```
    setcookie('embed_keyboards_no_locale_redirect','1',0, '/');
```

mcdurdin:
It is important to add the path for the cookie as shown above, but we still need your patch as well. I'll add the cookie as a separate PR

PR #739, opened 2026-05-14
refactor fallback locale 🌋

darcywong00:
Addresses review comment https://github.com/keymanapp/keyman.com/pull/733#issuecomment-4437784365 by taking a subset of those changes

> I think we want our URLs to redirect if a fallback is encountered, rather than accepting `es-Fooo-BA` as a substitute for `es`. And we don't want case insensitivity either; we don't want `es`, `eS`, `Es`, and `ES` all to return the same document, because URLs are case sensitive. So ... 

* Locale::setLocaleFromURL() changes
  * tweak the regex to be case-sensitive (capitalize script and region)
  * redirect to the fallback locale (if found)

This way, `es-Fooo-BA` and `es-ES` redirect to `es`.

Unspecified locale still redirects to `en`

Test-bot: skip

### DIFF
--- a/_includes/2020/templates/Head.php
+++ b/_includes/2020/templates/Head.php
@@ -39,10 +39,7 @@ class Head {
       // Redirect to /en/... if not a supported locale; this needs to be emitted
       // as a HTTP header before first content byte.
       if(Locale::invalidLocale()) {
-        if(preg_match('/^\\/[^\/]+\\/(.+)$/', $_SERVER['REQUEST_URI'], $matches)) {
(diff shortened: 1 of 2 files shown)


mcdurdin (changes requested):
After discussion, we opted to change this behaviour. Unmatched locales will not fallback but will 404. The only place we will do fallback is in app landing pages such as /go/platform/keyboard-downloads, where the set of locales in the app may be a superset of the set of locales in the website, and so we want to have a reasonable fallback behaviour there.

PR #750, opened 2026-05-20
feat: support lang parameter for keyboards embedded search

mcdurdin:
Test-bot: skip
Relates-to: keymanapp/keyman#15948

### DIFF
--- a/_content/keyboards/session.php
+++ b/_content/keyboards/session.php
@@ -1,8 +1,11 @@
 <?php
+  use \Keyman\Site\com\keyman\Locale;
+
   if(!isset($_SESSION)) {
     session_set_cookie_params(["SameSite" => "None"]);   // Allow use in iframe, needed for Download Keyboards dialog
     session_set_cookie_params(["Secure" => "true"]);     // None requires Secure to be set
     session_start();
+    $session_started = true;
   }
 
   if(isset($_REQUEST['embed'])) {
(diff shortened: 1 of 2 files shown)


_content/keyboards/session.php
```diff
@@ -42,6 +45,11 @@
     // and otherwise cookies are blocked
     setcookie('embed_keyboards_no_locale_redirect', '1', ["secure" => true, "samesite" => 'None', 'path' => '/']);
 
+    if(!isset($_SESSION['lang'])) {
+      $_SESSION['lang'] = isset($_REQUEST['lang']) ? $_REQUEST['lang'] : 'en';
+    }
```
darcywong00, line 50: What about the case where `$_SESSION['lang']` is already set (`fr`)
and a different `$_REQUEST['lang']` (`de`) comes in?

going from (appears in French)
http://keyman.com.localhost/fr/keyboards/?embed=none

to
http://keyman.com.localhost/go/android/14.0/download-keyboards?lang=de

appears in English

_includes/locale/Locale.php
```diff
@@ -79,19 +88,18 @@ private static function setLocaleFromURL() {
...
         $pageLocale = Locale::DEFAULT_LOCALE;
       }
-      self::setLocale($pageLocale);
+      self::setLocale($pageLocale, false);
     }
 
     /**
      * Set the current locales, with an array of fallbacks, ending in 'en'.
      * @param $locale - the new current locale (xx-YY as specified in crowdin %locale%)
      */
```
darcywong00, line 100: Can add a `@param $fallback`

_content/keyboards/session.php
mcdurdin: That can't happen in the embed mode, because it will be a new session if the dialog is re-opened.

darcywong00 (approved):
lgtm

PR #751, opened 2026-05-21
chore: Add km Crowdin strings for downloads

darcywong00:
This adds the Khmer Crowdin strings for the /downloads page

```
crowdin download -b master -l km
```

Since Crowdin escapes slashes when it downloads strings (`%1\$s` downloads as `%1\\$s`), I first started the site (init-container.sh) to fix the slashes before checking in the file.

http://keyman.com.localhost/km/downloads/

image

Test-bot: skip

### DIFF
--- a/_includes/locale/strings/downloads/km.php
+++ b/_includes/locale/strings/downloads/km.php
@@ -1,0 +1,93 @@
+<?php
+
+/*
+ * Keyman is copyright (C) SIL Global. MIT License.
+ *
+ * Khmer strings for downloads/index.php
+ * When exporting strings from crowdin, convert \\$s to \$s
+ */
+
+declare(strict_types=1);
+
+return [
+  # Downloads Page Title
+  "downloads_page_title" => "ការទាញយក Keyman",
+
+  # Downloads_page_description
+  "downloads_page_description" => "ការទាញយកជំនាន់ស្ថេរភាព​ Keyman",
+
+  # Get the latest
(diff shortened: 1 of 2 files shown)


_includes/locale/strings/downloads/km.php
```diff
@@ -0,0 +1,93 @@
...
+  # Keyman for macOS
+  "product_macos" => "Keyman សម្រាប់ macOS",
+
+  # Keyman for Android
+  "product_android" => "Keyman សម្រាប់ Android",
+
+  "available_on_play_store" => "Keyman សម្រាប់ Android ក៏មាននៅក្នុងកម្មវិធីទាញយក។",
+
+  # Keyman for iPhone and iPad
+  "product_ios" => "Keyman សម្រាប់ iPHone and iPad",
```
Meng-Heng:
```suggestion
  "product_ios" => "Keyman សម្រាប់ iPhone and iPad",
```
Hello, Darcy. I don't know when the H is capitalized. Now, I changed it to the correct "iPhone" in crowdin.

Meng-Heng (commented):
The translation strings for Khmer is LGTM

Image

_includes/locale/strings/downloads/km.php
darcywong00: Ah, should be "iPhone". I'll need to fix the English source string

ermshiperete (approved):

PR #752, opened 2026-05-21
chore: khmer translation now uses busra font

Meng-Heng:
This PR calls Busra font for Khmer translation (with Unicode range) on the website. Changing to this font allow easy skim reading, clearer and bolder characters, and readability. Please review.

Test-bot: skip

# User Interfaces

URL: http://keyman.com.localhost:8053/km/keyboards/
Screenshot 2026-05-21 at 4 24 45 in the afternoon

Screenshot 2026-05-21 at 4 24 58 in the afternoon

---

URL: http://keyman.com.localhost:8053/km/downloads/
image

image

### DIFF
--- a/cdn/dev/css/template.css
+++ b/cdn/dev/css/template.css
@@ -265,11 +265,7 @@ html {
 	unicode-range: U+1780-17FF, U+19E0-19FF;
 }
 
-/* Khmer */
-:lang(km) {
-	/* The browser will automatically use Busra for Khmer chars and Cabin or Sans-serif for others */
-	font-family: 'Busra','Cabin', sans-serif;
-}
+/* Alphabetically */
 
 /* Amharic */
 :lang(amh).lang-example {
@@ -289,6 +285,12 @@ html {
 /* Lao */
 :lang(lao).lang-example {
 	font-family: LaoWeb;
+}
+
+/* Khmer */
+:lang(km) {
+	/* The browser will automatically use Busra for Khmer chars and Cabin or Sans-serif for others */
+	font-family: Busra, Cabin, sans-serif;
 }
 
 /* Malayalam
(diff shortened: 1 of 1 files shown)


cdn/dev/css/template.css
```diff
@@ -257,6 +257,20 @@ html {
...
+	src: url('//s.keyman.com/font/deploy/Busra-Regular.ttf') format("truetype");
+	unicode-range: U+1780-17FF, U+19E0-19FF;
+}
+
+/* Khmer */
+:lang(km) {
+	/* The browser will automatically use Busra for Khmer chars and Cabin or Sans-serif for others */
+	font-family: 'Busra','Cabin', sans-serif;
+}
+
```
darcywong00: 1. For maintainability, can we move this alphabetically before Lao (line 289)?

2. I don't know CSS. Are Cabin and sans-serif already handled in line 173? (I note the other languages only reference one font for family, and not in quotes)

Meng-Heng: I'll try removing the two fonts to see if Busra is working properly or not. I added the Unicode range after font-family so this might fix it without having to specify three fonts.

Meng-Heng: I've got the answer now. The Cabin and sans-serif fonts are there as a fallback for other characters that aren't Khmer. 

Since I use `lang(km)` to recognize the `km` in the URL (`http://keyman.com.localhost:8053/km/keyboards/?q=k&x=0&y=0`) to trigger a font change, it changes the font for the entire web page. 

Busra font:
Screenshot 2026-05-22 at 5 02 50 in the afternoon

keyman.com:
Screenshot 2026-05-22 at 5 03 17 in the afternoon

Whereas the other `font-family` properties are being used for `.lang-example` selectors, so it does not impact any other elements; therefore, one value for `font-family` is enough.

darcywong00 (approved):
lgtm

PR #773, opened 2026-06-03
fix: update version landing pages with new i18n references

mcdurdin:
* Update the call to `downloadSection()` in version landing pages to match the new identifiers used in ui/downloads.php. Fixes the base issue for the versioned landing pages.

* Refactor `Locale::definePageScope` to declare standard `$_m_<domain>` variable and `_m_<domain>()` function, so that we can skip the Locale initialization boilerplate in more places. Also make it possible to declare multiple scopes (supporting includes in the future), and make the function idempotent.

* Update references to `[$]_m*` to the common pattern provided in `Locale::definePageScope`.

* Add version landing pages to _test/ so that they are crawled in the link checker.

Fixes: #764
Fixes: KEYMAN-COM-3H1
Test-bot: skip

### DIFF
--- a/_includes/locale/Locale.php
+++ b/_includes/locale/Locale.php
@@ -170,15 +170,44 @@
     }
 
     /**
-     * Defines a global variable for page locale strings and also
-     * tells locale system that current page uses locales
-     * @param $define -
-     * @param $id - folder containing locale strings, relative to /_includes/locale/strings
+     * Defines a global variable for page locale strings and also tells locale
+     * system that current page uses locales. Also defines a global function and
+     * variable for the page scope, so that we can use shorter function calls to
+     * get localized strings in the page code.
+     *
+     * @param $define - a string such as 'LOCALE_DOWNLOADS', which will be
+     *                  defined as a constant for the page scope
+     * @param $id - folder containing locale strings, relative to
+     *              /_includes/locale/strings, such as 'downloads'. This will be
+     *              used as the domain for the page scope, and also used to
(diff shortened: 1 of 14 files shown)


_includes/locale/Locale.php
```diff
@@ -170,15 +170,44 @@ public static function domain_js($domain) {
...
 
     /**
-     * Defines a global variable for page locale strings and also
-     * tells locale system that current page uses locales
-     * @param $define -
-     * @param $id - folder containing locale strings, relative to /_includes/locale/strings
+     * Defines a global variable for page locale strings and also tells locale
+     * system that current page uses locales. Also defines a global function and
+     * variable for the page scope, so that we can use shorter function calls to
+     * get localized strings in the page code.
```
ermshiperete, line 176: `$_m_Downloads` doesn't exactly seem to be shorter than `$_m` that we had before...

(not objecting this change, just the last sentence doesn't make completely sense for me)

ermshiperete (approved):

_includes/locale/Locale.php
mcdurdin: The 'shorter' here is in reference to the explicit call to `Locale::m`, not in reference to the `$_m()` / `_m()`: `Locale::m('LOCALE_DOWNLOADS', $id, $args);` 

It's also a little safer because it has a parse-time check for the function/variable name `_m_Downloads` rather than a runtime check for the parameter value `'LOCALE_DOWNLOADS'`, so we avoid a class of typo bugs.

PR #774, opened 2026-06-05
feat: select best language when redirecting to an i18nized page

mcdurdin:
Select the most appropriate language when a top-level page is encountered, either from the user's most recently chosen language, or else from the request Accept-Language header.

Install Negotiator to parse Accept-Language.

Fix a bug with the 404 page where the locale would be reset secretly by unexpected requests.

Remove the invalid locale redirect from Head.php because it cannot be encountered with the current htaccess patterns.

Establish the use of sessions throughout the site, to allow us to track the current language server-side. This could alternately be done with a cookie.

Fixes: #762
Fixes: #772
Test-bot: skip

### DIFF
--- a/_includes/locale/Locale.php
+++ b/_includes/locale/Locale.php
@@ -1,14 +1,15 @@
 <?php
-
-/*
- * Keyman is copyright (C) SIL Global. MIT License.
- */
-
+  /*
+  * Keyman is copyright (C) SIL Global. MIT License.
+  */
   declare(strict_types=1);
 
   namespace Keyman\Site\com\keyman;
 
   use \Keyman\Site\Common\KeymanHosts;
+  use \Keyman\Site\com\keyman\Session;
+
+  Session::Start();
 
   function define_display_locales() {
     $_defined_locales = json_decode(file_get_contents(__DIR__ . '/locales.json'), true);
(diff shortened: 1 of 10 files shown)


_includes/locale/Locale.php
```diff
@@ -31,6 +32,14 @@ class Locale {
 
     private static $invalidLocale = false;
 
+    /**
+     * For pages which don't belong in the _content hierarchy, such as 404.php,
+     * we do not want to save the selected locale to the session, because it may
+     * not be correct, and we also want to skip redirection when an invalid
+     * locale is found.
+     */
+    public static $saveLocale = true;
```
ermshiperete, line 41: The comment talks about not saving the locale yet we set `$saveLocale = true` ?

_includes/locale/Locale.php
```diff
@@ -42,6 +51,13 @@ public static function currentLocales() {
       return self::$currentLocales;
     }
 
+    /**
+     * Returns an array of available locales, e.g. ["en","de",...]
```
ermshiperete, line 55: Available locales here means locales that the server has, or locales that the client requests?

ermshiperete (approved):

_includes/locale/Locale.php
mcdurdin: OK, the default is to save locale, so we want to be able to turn that off on a per-page basis

_includes/locale/Locale.php
mcdurdin: Array of locales that the server can return -- those that the server has.

PR #775, opened 2026-06-05
fix: improve look and feel of globe icon on desktop layout

mcdurdin:
Adjusts the position, styling and interaction of the globe icon, on desktop devices. Does not change the mobile access, which will come in a separate PR.

Before:

image 
image

After:

image 
image

Relates-to: #771
Test-bot: skip


darcywong00 (approved):
Lgtm

PR #780, opened 2026-06-08
feat: add a banner where we don't have localized content

mcdurdin:
Two banners -- depending on context:

image

image

Fixes: #763
Test-bot: skip

### DIFF
--- a/_includes/2020/templates/Foot.php
+++ b/_includes/2020/templates/Foot.php
@@ -16,6 +16,9 @@
       $fields->pageLocale = Locale::pageLocale();
 ?>
 
+        <div id="locale-not-internationalized">This page has not yet been updated for languages other than English. <a href='/about/get-involved'>Can you help make this happen?</a></div>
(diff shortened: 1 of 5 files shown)


_includes/2020/templates/Foot.php
```diff
@@ -16,6 +16,9 @@ static function render(array $fields = []) {
       $fields->pageLocale = Locale::pageLocale();
 ?>
 
+        <div id="locale-not-internationalized">This page has not yet been updated for languages other than English. <a href='/about/get-involved'>Can you help make this happen?</a></div>
```
ermshiperete, line 19: Does this mean we show a banner at the top and at the bottom of the page?

If so we have duplicate ids.

ermshiperete (approved):

_includes/2020/templates/Foot.php
mcdurdin: I moved the banner to the bottom in the last commit, so there are no duplicates.

PR #786, opened 2026-06-10
feat: i18n of /windows

mcdurdin:
Add i18n of /windows/index.php. Other pages under /windows will come later.

Also, add a query param `LOCALE_DEBUG` so we can easily spot untranslated strings.

Fixes: #770
Test-bot: skip


_content/windows/index.php
```diff
@@ -2,172 +2,154 @@
...
     <a href="download"><img src="<?php echo cdn('img/download_button.png'); ?>" /></a>
 </div>
 
 <?php
-    if (betaTier()) {
-?>
-        <p>Want to try the Keyman for Windows <?php echo $beta_version ?> Beta? <a href="../beta/">Learn more</a></p>
-<?php
-    }
+  if (!betaTier()) {
```
ermshiperete: Isn't this the opposite of what it was before?


```suggestion
  if (betaTier()) {
```

ermshiperete (approved):

_content/windows/index.php
rc-swag: I think it is because it is only meant to offer the Beta version if they are already not on Beta.  Was it was wrong previously?

rc-swag (approved):

_content/windows/index.php
mcdurdin: sorry my bad, I was testing the changeover and definitely need to revert

ermshiperete (approved):

PR #787, opened 2026-06-12
chore: use common function for backend http requests

mcdurdin:
Establish Util::call_api_keyman_com and Util::call_downloads_keyman_com as common functions for making http requests to these servers. Use curl instead of file_get_contents as it is more reliable and we have more control over parameters.

The pattern for now could stand to be refactored further in the future to provide API functions for each endpoint and reduce repetition and fragility.

Test-bot: skip


_includes/2020/Util.php
```diff
@@ -32,4 +32,37 @@ static function cdn($file) {
...
+
+    private static function call_keyman_site($site, $url, $testFixture) {
+      if(KeymanHosts::Instance()->Tier() == KeymanHosts::TIER_TEST) {
+        return file_get_contents(__DIR__ . '/../../tests/fixtures/' . $testFixture);
+      }
+
+      // curl library is more reliable than file_get_contents
+      $curl_handle=curl_init();
+      curl_setopt($curl_handle, CURLOPT_URL, $site . $url);
+      curl_setopt($curl_handle, CURLOPT_CONNECTTIMEOUT, 2);
```
ermshiperete, line 44: Do we have to set a general timeout as well (otherwise the default applies which is 0, i.e. no timeout)?

```suggestion
      curl_setopt($curl_handle, CURLOPT_CONNECTTIMEOUT, 2);
      curl_setopt($curl_handle, CURLOPT_TIMEOUT, 60);
```

ermshiperete (approved):

_includes/2020/Util.php
mcdurdin: Probably? We should make it 15 seconds or something like that? 60 seconds seems too long to me for an API request.

mcdurdin:
```suggestion
      curl_setopt($curl_handle, CURLOPT_CONNECTTIMEOUT, 2);
      curl_setopt($curl_handle, CURLOPT_TIMEOUT, 15);      
```

mcdurdin: Hmm, maybe not. If we request _all_ keyboards, that could take longer than 15 seconds if server is loaded...

mcdurdin:
```suggestion
      curl_setopt($curl_handle, CURLOPT_CONNECTTIMEOUT, 2);
      curl_setopt($curl_handle, CURLOPT_TIMEOUT, 60);
```

_includes/2020/Util.php
```diff
@@ -32,4 +32,39 @@ static function cdn($file) {
...
+      curl_setopt($curl_handle, CURLOPT_TIMEOUT, 60);
+      curl_setopt($curl_handle, CURLOPT_TIMEOUT, 15);
```
mcdurdin:
```suggestion
```

_includes/2020/Util.php
ermshiperete: It seems 60s was what we (implicitly) had with the old code.

PR #793, opened 2026-06-16
chore: use keymancom instead of keyman for localization site

mcdurdin:
Content on websites is in the keymancom project on Crowdin.

Test-bot: skip


darcywong00 (approved):
lgtm

PR #794, opened 2026-06-16
chore: verify bootstrap version format before merge

mcdurdin:
It is important that the bootstrap version is pointing to a published tag, so that merged changes do not end up pointing to an ephemeral ref on the shared-sites repository. This change splits the bootstrap version into a separate file resources/.bootstrap-required-version, so that this version string can be easily checked in ci.yml, preventing merge of pull requests that point to branch names instead of tag names. For simplicity, the test verifies the shape of the version string as 'v#.#.#', but does not verify that it actually points to a tag, so in theory it could be fooled, but in practice we never use that pattern for branch names.

Test-bot: skip

### DIFF
--- a/resources/.bootstrap-required-version
+++ b/resources/.bootstrap-required-version
@@ -1,0 +1,1 @@
+feat/linkinator-and-central-test-script
--- a/resources/.bootstrap-required-version
+++ b/resources/.bootstrap-required-version
@@ -1,1 +1,1 @@
-feat/linkinator-and-central-test-script
+v1.0.14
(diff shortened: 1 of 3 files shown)


resources/.bootstrap-required-version
```diff
@@ -0,0 +1 @@
+feat/linkinator-and-central-test-script
```
darcywong00: When does this get updated?
(and will it need to be a v#.#.#?)

mcdurdin: That's what this PR is testing, and yeah, because I accidentally merged this morning before updating to a tag!

mcdurdin: So this PR will stop that mistake happening again -- CI / build-test will verify it

ermshiperete (approved):

PR #802, opened 2026-06-24
feat: show keyboard status in Details

### LINKED ISSUE #566: Feature: Indicate Keyboard Status [Stable|Experimental|Legacy]
I hope this is the correct place to post a thought for a website refhinement.  On a keyboard page (for example, https://keyman.com/keyboards/gff_sbs) it would be beneficial to users if the keyboard "status" (for lack of a better term) were indicated, with some note given so they are aware that the keyboard package may have future compatibility issues.

For example, in the keyman keyboards repo, keyboards are divided into 3 categories "Release", "Legacy" and "Experimental", which is helpful for developers.  However, this same information is not available to users at download time. A simple adjustment would be to add a "Status" field, or maybe a "Compatibility" field under the "Keyboard Details" section that presented terms like "Stable", "Legacy" and "Experimental".  Then a hyperlink or tooltip balloon could appear that explains what the term means (or it could be a sentence following the term).  

- **Stable:** Safe to use. This keyboard is actively maintained and complies with international standards such as Unicode.
- **Legacy:** Working but unsupported. This keyboard is no longer supported by its maintainer.  It may stop working on newer versions of supported platforms (Windows, MacOS, etc.) and defects are unlikely to be fixed.
-  **Experimental:** Use at your own risk. This keyboard may be considered an experiment by its maintainer.  The typing method is subject to change without notice. Fonts packaged with the keyboard may also change, making previously typed work incompatible (unreadable) with updates.

The "Source" field of the "Keyboard Details" section does indicate a path with this github categorization, but I think this is still too subtle for users and does not communicate the implications for what the user is getting into by installing the keyboard.

mcdurdin:
Add a keyboard status field to the details, and an explanation. Extend the UI with a tooltip explanation of the field, exposed with dotted underline. Note that the tooltip will not be accessible on touch devices at this time (both this and other tooltips on the page are areas for future enhancement).

Fixes: #566
Test-bot: skip

### DIFF
--- a/_includes/includes/ui/keyboard-details.php
+++ b/_includes/includes/ui/keyboard-details.php
@@ -276,20 +276,24 @@ END;
 
         // Package Status
 
-        $status = 'unknown';
-        if(self::$keyboard->sourcePath) {
-          $sourcePath = self::$keyboard->sourcePath;
-          if(preg_match('/^experimental/', $sourcePath)) {
-            $status = 'experimental';
-          } else if(preg_match('/^legacy/', $sourcePath)) {
-            $status = 'legacy';
-          } else if(preg_match('/^release/', $sourcePath)) {
-            $status = 'stable';
-          }
-        }
+        $status = self::GetStatusFromSourcePath(self::$keyboard->sourcePath ?? '');
         self::$packageStatus = htmlentities($_m_Keyboards_Details("status_$status"));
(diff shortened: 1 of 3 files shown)


sgschantz (commented):
Looks good to me, but can you break out the logic for setting the status property into a separate function instead of making that already large function even larger?

sgschantz (approved):

